PR #24, opened 2021-12-27
Allow ignoring go list errors

sverdlov93:
-----

In some cases, we see that running `go list` on some Go packages may fail.
We should allow ignoring the errors in such cases and build the Go dependency tree, even if partial.


utils/goutils.go
```diff
@@ -140,7 +140,12 @@ func GetDependenciesList(projectDir string, log Log) (map[string]bool, error) {
 	}
 	output, err := runDependenciesCmd(projectDir, append(cmdArgs, "-f", "{{with .Module}}{{.Path}}@{{.Version}}{{end}}", "all"), log)
 	if err != nil {
-		return nil, err
+		// Errors occurred during running "go list". Run again and this time ignore errors (with '-e')
```
eyalbe4, line 143: during --> while

eyalbe4 (changes requested):

PR #46, opened 2022-01-08
Improve build-info comparison

Or-Geva:
-----


utils/compare/compare.go
```diff
@@ -0,0 +1,39 @@
+package compare
```
eyalbe4, line 1: I'm not sure we really need a new package for a single file.

utils/compare/compare.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"reflect"
+	"sort"
+)
+
+// Check for equality on slices without order.
+func IsEqualSlices(a, b []string) bool {
```
eyalbe4, line 9: Let's add a unit test for this function.

utils/compare/compare.go
```diff
@@ -0,0 +1,39 @@
...
+	copy(b_copy, b)
+
+	sort.Strings(a_copy)
+	sort.Strings(b_copy)
+
+	return reflect.DeepEqual(a_copy, b_copy)
+}
+
+func IsEqual2DSlices(a, b [][]string) bool {
+	return IsEqualSlices(to1DSlice(a), to1DSlice(b))
```
eyalbe4, line 27: Let's add a unit test for this function.

eyalbe4 (changes requested):

utils/compare/compare.go
Or-Geva: Ok, but I had to move npm utils to the build utils because otherwise, we will have an import cycle.

eyalbe4: Got it. In that case, let's leave it as is,

eyalbe4 (approved):

eyalbe4 (approved):

PR #49, opened 2022-01-11
Remove pointer from  embedded checksum struct

Or-Geva:
-----
The struct `entities.Checksum` located in a few places as an embedded field.
It's bad practice to use it as a pointer and not a struct for a couple of reasons:

First, most of the time, the wrapper struct, which  `entities.Checksum` is embedded inside, is a pointer itself, so there is no real benefit for using nested pointers.
Seconds, `entities.Checksum` fields are string, when referencing one of the fields, it looks like no nil check is required but `entities.Checksum` itself could be nil.
This may open the door for runtime exceptions such as here https://github.com/jfrog/jfrog-cli/issues/1297.

The solution to this problem is simple, use struct instead of pointer as embedded field.
Related PRs:
* https://github.com/jfrog/jfrog-cli-core/pull/306
* https://github.com/jfrog/jfrog-client-go/pull/504

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -340,7 +340,7 @@ type partialModule struct {
 	moduleType   entities.ModuleType
 	artifacts    map[string]entities.Artifact
 	dependencies map[string]entities.Dependency
-	checksum     *entities.Checksum
+	checksum     entities.Checksum
 }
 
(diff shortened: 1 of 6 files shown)


yahavi (approved):
LGTM

PR #51, opened 2022-01-17
Fix redundant module for env or vcs

RobiNino:
-----
Following jfrog/jfrog-cli#1349

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -351,7 +351,9 @@ func extractBuildInfoData(partials entities.Partials) ([]entities.Module, entiti
 	issuesMap := make(map[string]*entities.AffectedIssue)
 	for _, partial := range partials {
 		moduleId := partial.ModuleId
-		addNewPartialModule(partial, partialModules)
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -391,6 +389,18 @@ func extractBuildInfoData(partials entities.Partials) ([]entities.Module, entiti
...
 
+func addNewPartialModule(partial *entities.Partial, partialModules map[string]*partialModule) {
+	// Skip if module ID already exists.
+	if partialModules[partial.ModuleId] != nil {
+		return
+	}
+	// Skip if module is of Env or Vcs.
+	if partial.ModuleType == "" && (partial.Env != nil || partial.VcsList != nil) {
+		return
+	}
```
eyalbe4, line 400: I suggest changing the above code - 
```
	if partial.ModuleType == "" && (partial.Env != nil || partial.VcsList != nil) {
		return
	}
```
to - 
```
	if partial.ModuleType == "" {
		return
	}
```

eyalbe4 (changes requested):

PR #69, opened 2022-03-14
Bugix - Skip optional npm dependencies

Or-Geva:
-----
Projects with Npm dependencies may have optional dependencies. We should skip adding those dependencies to the buildinfo.json in case they aren't available.


eyalbe4 (changes requested):
In the debug log message added by the ```printMissingDependenciesWarning``` function, I suggest removing the new-line  (```\n```). There's no reason to cut log messages into two lines.

PR #76, opened 2022-04-10
Add build-info comparison using regex

Or-Geva:
-----


entities/buildinfo.go
```diff
@@ -491,8 +566,21 @@ func (c *Checksum) IsEmpty() bool {
 	return c.Md5 == "" && c.Sha1 == "" && c.Sha256 == ""
 }
 
-func (c *Checksum) IsEqual(other Checksum) bool {
-	return c.Md5 == other.Md5 && c.Sha1 == other.Sha1 && c.Sha256 == other.Sha256
+func (c *Checksum) IsEqual(expected Checksum) (bool, error) {
```
eyalbe4, line 569: Shouldn't we also add a description for this function, explaining that the expected checksum regex can also include a regex?

expected --> other
Tthe above rename is also relevent for other "euqal" functions in this file. It makes more sense to use "expected" for test functions, and not comparison functions.

eyalbe4 (changes requested):

PR #108, opened 2022-10-03
Unite python3 code with cli-core

sverdlov93:
-----

### DIFF
--- a/utils/pythonutils/piputils.go
+++ b/utils/pythonutils/piputils.go
@@ -3,6 +3,7 @@ package pythonutils
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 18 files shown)


utils/pythonutils/piputils.go
```diff
@@ -142,14 +140,14 @@ func getEgginfoPkginfoContent(setuppyFilePath string) (output []byte, err error)
 	return extractPackageNameFromEggBase(eggBase)
 }
 
-func getPythonExecutable() (string, string) {
+func GetPython3Executable() (string, string) {
 	windowsPyArg := ""
-	pythonExecutable, pathErr := exec.LookPath("python3")
-	if pathErr != nil || pythonExecutable == "" {
+	pythonExecutable, _ := exec.LookPath("python3")
```
eyalbe4, line 145: I understand why you're ignoring the error returned by this function now, but I think we can achieve the goal of trying a few python option, without swallowing the error completely. See my suggestion for function's code below. Let me know if you think - 
```
func GetPython3Executable() (pythonExecutable string, windowsPyArg string, err error) {
	pythonExecutable, err = exec.LookPath("python3")
	if pythonExecutable == "" {
		if runtime.GOOS == "windows" {
			// If the OS is Windows try using Py Launcher: 'py -3'
			pythonExecutable, err = exec.LookPath("py")
			if pythonExecutable != "" {
				windowsPyArg = "-3"
			}
		}
		// Try using 'python' if 'python3'/'py' couldn't be found
		if pythonExecutable == "" {
			pythonExecutable = "python"
		}
	}
	return
}
```

eyalbe4 (changes requested):

PR #123, opened 2023-01-19
Add static analysis to test files

RobiNino:
-----

### DIFF
--- a/utils/goutils_test.go
+++ b/utils/goutils_test.go
@@ -122,9 +122,7 @@ func testGetDependenciesList(t *testing.T, testDir string) {
 	// Since Go 1.16 'go list' command won't automatically update go.mod and go.sum.
 	// Check that we roll back changes properly.
 	newSumFileContent, _, err := getGoSum(goModPath, log)
-	if !reflect.DeepEqual(originSumFileContent, newSumFileContent) {
-		t.Errorf("go.sum has been modified and didn't rollback properly")
-	}
+	assert.Equal(t, newSumFileContent, originSumFileContent, "go.sum has been modified and didn't rollback properly")
 
 	expected := map[string]bool{
 		"golang.org/x/text:v0.3.3": true,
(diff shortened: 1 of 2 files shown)


utils/goutils_test.go
```diff
@@ -41,71 +41,47 @@ rsc.io/sampler:v1.3.0
...
 func TestGetProjectRoot(t *testing.T) {
 	wd, err := os.Getwd()
-	if err != nil {
-		t.Error(err)
-	}
-	defer os.Chdir(wd)
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, os.Chdir(wd))
+	}()
```
yahavi, line 47:
```suggestion
	defer assert.NoError(t, os.Chdir(wd))
```

utils/goutils_test.go
```diff
@@ -41,71 +41,47 @@ rsc.io/sampler:v1.3.0
...
-	}
+	assert.NoError(t, err)
+	assert.Equal(t, projectRoot, root)
 
 	// CD back to the original directory.
-	if err := os.Chdir(wd); err != nil {
-		t.Error(err)
+	if !assert.NoError(t, os.Chdir(wd)) {
+		return
 	}
```
yahavi, line 65: I think that flat code in tests is easier to understand
```suggestion
	assert.NoError(t, os.Chdir(wd))
```

utils/goutils_test.go
```diff
@@ -120,34 +96,32 @@ func TestGetDependenciesListWithIgnoreErrors(t *testing.T) {
...
-	}
+	actual, err := GetDependenciesList(filepath.Join(goModPath), log)
+	assert.NoError(t, err)
 
 	// Since Go 1.16 'go list' command won't automatically update go.mod and go.sum.
-	// Check that we rollback changes properly.
-	newSumFileContent, _, err := getGoSum(gomodPath, log)
+	// Check that we roll back changes properly.
+	newSumFileContent, _, err := getGoSum(goModPath, log)
 	if !reflect.DeepEqual(originSumFileContent, newSumFileContent) {
```
yahavi, line 125: Let's remove the reflection here. We can use a simple assert.Equals with the text below as the message.

yahavi (changes requested):

utils/goutils_test.go
RobiNino: Why? we change back to the wd as a form of cleanup.

utils/goutils_test.go
RobiNino: It is easier to understand, but if changing dir failed here the rest of the test is irrelevant. This is used to fail the test fast

Issue #161, opened 2023-05-12
Support overriding the version of the CLI extractors to use


markrekveld:
**Is your feature request related to a problem? Please describe.**
It's impossible to test PRs for extractors without also building a custom CLI and using this. But then, what are you testing? Any issue during the CLI build to validate and test the extractor build makes it harder to do a proper job.

Take this case:
I found a (in my eyes) bug in the extractor implementation for Maven. It introduces a 'dummy' repository but this is causing numerous issues with Maven plugins that cannot take this kind of repository into account.

So I opened a PR in an attempt to resolve this: https://github.com/jfrog/build-info/pull/683
But this introduced a bug and needed to be reverted.
This is not only annoying for me to have to revisit the PR again, but also other customers affected by the introduced bug.
It was easy to resolve and a new PR was opened to address the bug introduced https://github.com/jfrog/build-info/pull/715

I'm confident that the bug would not have been introduced if it was easier to test the custom version of the extractor.

**Describe the solution you'd like to see**
I would like to be able to override the version of the extractor used during Maven builds so I can use or test a custom version of the extractor.

Currently you can provide a different remote server to download the extractor from, but the version is hardcoded at https://github.com/jfrog/build-info-go/blob/a6ed2e89a5bf16feff10be3333c5f69821b0b903/build/maven.go#LL21C29-L21C29
But it would be great if this version can be overridden by an environment variable just like the remote server can be overridden.

**Describe alternatives you've considered**
The only alternative is to also build a custom CLI version to test the custom extractor version, which is not really an alternative if you ask me.

**Additional context**
See also opened [support ticket](https://support.jfrog.com/s/tickets/5006900005xKBfl/support-overiding-the-buildinfoextractor-version-used-by-the-jfrog-cli-mvn-command)

eyalbe4:
Hi @markrekveld,
First of all - thanks a lot for being active and helping us improve the code. This is highly appreciated!
As for your request, we do not believe that the approach of overriding the versions of internal dependencies is the right approach, because this may lead to compatibility issues. For development purposes, you can always modify the code to use a specific version.

markrekveld:
Hi @eyalbe4 

Thanks for your reply.

I'm a but lost. The CLI is hardcoded to a specific version of the build info extractor. How can I use the CLI with a different version of the extractor to test a pull request like the one I linked?

[...]

PR #168, opened 2023-06-15
Allow overriding of the maven extractor version that is downloaded

markrekveld:
My attempt of implementing #161

### DIFF
--- a/build/maven.go
+++ b/build/maven.go
@@ -15,12 +15,12 @@ import (
 )
 
 const (
-	MavenHome                       = "M2_HOME"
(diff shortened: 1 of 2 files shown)


markrekveld:
If this gets accepted then this line `return filepath.Join(dependenciesPath, "maven", build.MavenExtractorDependencyVersion), nil` will need to be changed to `return filepath.Join(dependenciesPath, "maven", build.MavenExtractorDependencyVersion()), nil` in https://github.com/jfrog/jfrog-cli-core/blob/master/utils/mvn/utils.go#L122

PR #172, opened 2023-07-26
Build Dependency Tree - Yarn V1

eranturgeman:
-----

Added support for building dependency tree for Yarn project in version lesser than 2.0.0
building a dependency tree is now supported for all Yarn versions
new tests and test suitcases were added. old tests were added with extra checks


eranturgeman:

build/utils/yarn.go
```diff
@@ -91,12 +94,19 @@ func GetYarnExecutable() (string, error) {
...
-// Note that in some versions of Yarn, the version of the root package is '0.0.0-use.local', instead of the version in the package.json file.
 func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInfo, log utils.Log) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
-	// Run 'yarn info'
-	responseStr, errStr, err := runYarnInfo(executablePath, srcPath)
+	executableVersion, err := getVersion(executablePath, srcPath)
+	if err != nil {
+		return
+	}
+
+	isV2AndAbove := strings.Compare(executableVersion, YarnV2Version) >= 0
```
sverdlov93, line 106: you can't compare versions with strings.compare.
here is an example why not - `isV2AndAbove := strings.Compare("11.0.0", "2.0.0") >= 0`
you should check our gofrog func (v *Version) Compare(ver1 string) function
or just to take the first number on the version string `"x.y.z" > x` , and just see if its more then 2.

build/utils/npm.go
```diff
@@ -410,9 +410,13 @@ func GetNpmVersionAndExecPath(log utils.Log) (*version.Version, string, error) {
 }
 
 type PackageInfo struct {
-	Name    string `json:"name,omitempty"`
-	Version string `json:"version,omitempty"`
-	Scope   string
+	Name                 string            `json:"name,omitempty"`
+	Version              string            `json:"version,omitempty"`
+	Dependencies         map[string]string `json:"dependencies"`
```
sverdlov93, line 415: omitempty?

build/utils/npm.go
```diff
@@ -410,9 +410,13 @@ func GetNpmVersionAndExecPath(log utils.Log) (*version.Version, string, error) {
...
+	DevDependencies      map[string]string `json:"devDependencies"`
+	PeerDependencies     map[string]string `json:"peerDependencies"`
+	OptionalDependencies map[string]string `json:"optionalDependencies"`
+	Scope                string
```
sverdlov93, line 419: why no json field?

build/utils/yarn.go
sverdlov93: having said that, why not just use - validateYarnVersion same as today but without returning an error.

build/utils/yarn.go
```diff
@@ -108,9 +118,71 @@ func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInf
...
+
+// buildYarnV1DependencyMap builds a map of dependencies for Yarn versions < 2.0.0
+// Pay attention that in Yarn < 2.0.0 the project itself with its direct dependencies is not presented when running the
+// command 'yarn list' therefore the root is built manually.
+func buildYarnV1DependencyMap(packageInfo *PackageInfo, responseStr string) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
+	dependenciesMap = make(map[string]*YarnDependency)
+	var depTree Yarn1Data
+	err = json.Unmarshal([]byte(responseStr), &depTree)
+	if err != nil {
+		return
```
sverdlov93, line 154: and error that will return here will be very confusing to the user, we should wrap it with an explaining error:
err = errors.new("couldn't parse yarn list results etc etc: " + err)

build/utils/yarn.go
```diff
@@ -108,9 +118,71 @@ func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInf
...
+	}
+
+	locatorsMap := make(map[string]string)
+	for _, curDependency := range depTree.Data.DepTree {
```
sverdlov93, line 158: let's validate that deptree.data is no nil because if it is, this line will throw an ugly panic

build/utils/yarn.go
```diff
@@ -108,9 +118,71 @@ func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInf
...
+		locatorsMap[curDependency.Name[:strings.Index(curDependency.Name[1:], "@")+1]] = curDependency.Name
```
sverdlov93, line 159: let's discuss all this index and +1 +2 conditions

sverdlov93 (commented):
not reviewed all PR yet

build/yarn_test.go
```diff
@@ -131,16 +131,24 @@ func TestCollectDepsForYarnProjectWithTraverse(t *testing.T) {
 	buildInfo, err := yarnBuild.ToBuildInfo()
 	assert.NoError(t, err)
 	assert.Len(t, buildInfo.Modules, 1)
-	validateModule(t, buildInfo.Modules[0], 1, 0, "build-info-go-tests:v1.0.0", entities.Npm, true)
-	assert.Equal(t, "json:9.0.6", buildInfo.Modules[0].Dependencies[0].Id)
-	assert.Equal(t, "test123", buildInfo.Modules[0].Dependencies[0].Sha1)
+	validateModule(t, buildInfo.Modules[0], 4, 0, "build-info-go-tests:v1.0.0", entities.Npm, true)
+	dependencies := []string{
```
eranturgeman, line 135: I switched the checking logic since we have 4 dependencies now instead of 1, and since they are kept in a map I cannot know the order in which they will be traversed. therefore I validated the correct amout of dependencies and checked if we have each of the dependencies I expect to have

build/utils/yarn.go
eranturgeman: I tried to use validateYarnVersion
validateYarnVersion is located in build-info-go/build/yarn.go and the file I need it for is build-info-go/build/utils/yarn.go which is being imported by build-info-go/build/yarn.go. this creates a circular import and will not compile

build/utils/yarn.go
eranturgeman: since depTree.Data is not a pointer, and It is declared above (therefore I can access those structs fields even before the unmarshal) there is no need to check for nil value.
But I added this check for depTree.Data.DepTree

build/utils/npm.go
eranturgeman: there is no Scope field in package.json files therefore the Scope field is not intended for the json parsing

build/utils/yarn_test.go
```diff
@@ -24,24 +25,76 @@ func TestGetYarnDependencyKeyFromLocator(t *testing.T) {
...
+	for _, dependency := range root.Details.Dependencies {
+		assert.Contains(t, expectedLocators, dependency.Locator)
+	}
+
+	// checking dependencyMap
+	// NOTICE: the test uses fixed package versions. if for some reason dependencies in those packages (with those versions) changes
+	// (i.e. new dependencies added) make sure to fix the checks below
+	assert.Len(t, dependenciesMap, 6)
+	for key, val := range dependenciesMap {
+		switch {
```
sverdlov93, line 75: you can split the key to get the prefixed string and then run a simple` switch key { ... }`

build/utils/yarn.go
```diff
@@ -91,26 +93,105 @@ func GetYarnExecutable() (string, error) {
...
+
+// buildYarnV1DependencyMap builds a map of dependencies for Yarn versions < 2.0.0
+// Pay attention that in Yarn < 2.0.0 the project itself with its direct dependencies is not presented when running the
+// command 'yarn list' therefore the root is built manually.
+func buildYarnV1DependencyMap(packageInfo *PackageInfo, responseStr string) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
+	dependenciesMap = make(map[string]*YarnDependency)
+	var depTree Yarn1Data
+	err = json.Unmarshal([]byte(responseStr), &depTree)
+	if err != nil {
+		err = errors.New("Couldn't parse 'yarn list' results in order to create the dependencyMap:\n" + err.Error())
```
sverdlov93, line 159:
```suggestion
		err = errors.New("couldn't parse 'yarn list' results in order to create the dependencyMap:\n" + err.Error())
```

build/utils/yarn.go
```diff
@@ -91,26 +93,105 @@ func GetYarnExecutable() (string, error) {
...
+		return
+	}
+
+	if depTree.Data.DepTree == nil {
+		err = errors.New("Error: (buildYarnV1DependencyMap) depTree struct received a nil value in a required field: depTree.Data.DepTree")
```
sverdlov93, line 164: should start with lowercase
no need to write `error`
the error is not user friednly

build/utils/yarn.go
```diff
@@ -180,8 +313,12 @@ type YarnDependency struct {
 
 func (yd *YarnDependency) Name() string {
```
sverdlov93, line 314: can be replaced with:
strings.replace( value ,  "@npm:" , "@")

build/utils/yarn_test.go
```diff
@@ -24,24 +28,157 @@ func TestGetYarnDependencyKeyFromLocator(t *testing.T) {
...
+	defer func() {
+		assert.NoError(t, yarnLockFile.Close())
+	}()
+	err = yarnLockFile.Truncate(0) // This line erases the file's content without deleting the file itself
+	assert.NoError(t, err, "Could not erase yarn.lock file content")
+
+	pacInfo := PackageInfo{Name: "build-info-go-tests"}
+	_, _, err = GetYarnDependencies(executablePath, projectSrcPath, &pacInfo, &utils.NullLog{})
+	assert.Error(t, err)
+	assert.True(t, strings.Contains(err.Error(), "fetching dependencies failed since '"+pacInfo.Name+"' doesn't present in your lockfile\nPlease run 'yarn install' to update lockfile\n"))
```
sverdlov93, line 74:
```suggestion
	assert.Contains(t, err.Error(), "fetching dependencies failed since '"+pacInfo.Name+"' doesn't present in your lockfile\nPlease run 'yarn install' to update lockfile\n")
```

build/utils/yarn_test.go
```diff
@@ -24,24 +28,157 @@ func TestGetYarnDependencyKeyFromLocator(t *testing.T) {
...
+			}
+		case "xml":
+			assert.Equal(t, "1.0.1", depInfo.Details.Version)
+			assert.Nil(t, depInfo.Details.Dependencies)
+		case "json":
+			assert.Equal(t, "9.0.6", depInfo.Details.Version)
+			assert.Nil(t, depInfo.Details.Dependencies)
+		case "loose-envify":
+			assert.NotNil(t, depInfo.Details.Dependencies)
+			assert.Equal(t, len(depInfo.Details.Dependencies), 1)
```
sverdlov93, line 155:
```suggestion
			assert.Len(t, depInfo.Details.Dependencies, 1)
```

sverdlov93: ?

sverdlov93 (approved):

PR #226, opened 2023-12-27
Fix Nuget nuspec files UTF-16 encoding issue

sverdlov93:
-----
Sometimes the nuspec file is wrongly encoded in utf-16, the actual encoding is utf-8 but the xml header has `enocding="utf-16"` key.
`xml.Unmarshal` doesn't support utf-16 encoding, so we need to convert the header to utf-8.

should fix https://github.com/jfrog/jfrog-cli/issues/840 and https://github.com/jfrog/jfrog-azure-devops-extension/issues/464


build/utils/dotnet/dependencies/packagesconfig.go
```diff
@@ -218,9 +218,11 @@ func createNugetPackage(packagesPath string, nuget xmlPackage, nPackage *nugetPa
 	if err != nil {
 		return nil, err
 	}
-
+	// Sometimes the nuspec file is wrongly encoded in utf-16, the actual encoding is utf-8 but the xml header has 'enocding="utf-16"' key.
+	// xml.Unmarshal doesn't support utf-16 encoding, so we need to convert the header to utf-8.
+	utf8String := strings.Replace(string(nuspecContent), "utf-16", "utf-8", 1)
```
yahavi, line 223: I think that keeping the content in bytes format should be more efficient:
```suggestion
	utf8Bytes := bytes.Replace(nuspecContent, []byte("utf-16"), []byte("utf-8"), 1)
```
You can even go further and put []byte("utf-16") and []byte("utf-8") in class consts.

yahavi (changes requested):

yahavi (approved):
LGTM

PR #233, opened 2024-02-18
Add exclude pattern for dotnet

attiasas:
-----

Add an option to exclude patterns for dotnet dependencies.


build/utils/dotnet/solution/solution.go
```diff
@@ -174,14 +174,27 @@ func (solution *solution) loadProjects(slnProjects []project.Project, log utils.
...
+func (solution *solution) parseProjectsFromSolutionFile(slnProjects []string, excludePattern string, log utils.Log) ([]project.Project, error) {
 	var projects []project.Project
 	for _, projectLine := range slnProjects {
 		projectName, projFilePath, err := parseProjectLine(projectLine, solution.path)
 		if err != nil {
 			log.Error(err)
 			continue
 		}
+		// Exclude projects by pattern.
+		if len(excludePattern) > 0 {
```
eranturgeman, line 186: consider putting this logic into a function like "isPatternExcluded) or something for the clarity of the flow

build/utils/dotnet/solution/solution_test.go
```diff
@@ -160,7 +160,7 @@ func TestLoad(t *testing.T) {
 	// 1. located in the project's root directory.
 	// 2. located in solutions directory.
 	solution := solution{path: filepath.Join(wd, "testdata", "nugetproj", "solutions"), slnFile: "nugetproj.sln"}
-	solutions, err := Load(solution.path, solution.slnFile, log)
+	solutions, err := Load(solution.path, solution.slnFile, "", log)
```
eranturgeman, line 163: in this test we can add a testcase for Load with an excluded pattern

eranturgeman (commented):
Some fixes suggested

build/utils/dotnet/solution/solution.go
```diff
@@ -174,14 +174,27 @@ func (solution *solution) loadProjects(slnProjects []project.Project, log utils.
...
+			log.Debug(fmt.Sprintf("Testing to exclude projects by pattern: %s", excludePattern))
+			exclude, err := regexp.MatchString(excludePattern, projFilePath)
+			if err != nil {
+				log.Error(err)
+				continue
```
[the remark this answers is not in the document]
attiasas: I think an error is OK in this case, we log it just like we log when we can't parse at the above line.
There could be an error from the pattern or a lot of other reasons...
We got an error for this project so we report it and skip to the next, we don't want to fail all the build but still an error occur and it is not the full one, the customer may be at fault since we rely on input given to us.

build/utils/dotnet/solution/solution_test.go
attiasas: tests added

PR #246, opened 2024-04-07
Ensure `CreateTempDir()` is Multi-Process Safe

noyshabtay:
-----
Refining `CreateTempDir()` by appending a wildcard `*` at the end of the `MkdirTemp()` pattern argument. This ensures that a unique random string replaces the last `*`, preventing concurrent processes or goroutines from selecting the same directory. This approach is preferred over using only `strconv.FormatInt(time.Now().Unix(), 10)`, which may not offer sufficient distinction between processes.

### DIFF
--- a/utils/fileutils_test.go
+++ b/utils/fileutils_test.go
@@ -56,17 +56,17 @@ func TestReadNLines(t *testing.T) {
 }
 
 func TestCreateTempDir(t *testing.T) {
-    tempDir, err := CreateTempDir()
-    assert.NoError(t, err)
+	tempDir, err := CreateTempDir()
+	assert.NoError(t, err)
 
-    _, err = os.Stat(tempDir)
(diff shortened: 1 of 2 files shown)


utils/fileutils_test.go
```diff
@@ -54,3 +54,17 @@ func TestReadNLines(t *testing.T) {
...
+    assert.NoError(t, err)
+
+    _, err = os.Stat(tempDir)
+	assert.NotErrorIs(t, err, os.ErrNotExist)
+
+	// Check that a timestamp can be extracted from the temp dir name
+	_, err = extractTimestamp(tempDir)
+	assert.NoError(t, err)
+
+	assert.NoError(t, os.RemoveAll(tempDir))
```
omerzi, line 69: It think it is better to put it in a `defer` after the directory creation

PR #255, opened 2024-05-23
Allow the `GetDependenciesList` for Go to accept an error handler

asafambar:
…n when building go tree

- [X] All [tests](https://github.com/jfrog/build-info-go#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [X] All [static analysis checks](https://github.com/jfrog/build-info-go/actions/workflows/analysis.yml) passed.
- [X] This pull request is on the dev branch.
- [X] I used gofmt for formatting the code before submitting the pull request.
-----

The PR introduces support for recognizing errors that occur while building the Go dependency tree. The primary reason for adding this support is to detect errors originating from the curation service. It's essential to catch these errors during the tree-building process, as they can occur when fetching dependencies from Artifactory as a result of running the go list command. The code changes involve passing a function to handle these errors, allowing the libraries that use it to manage error handling and reduce code duplication.

### DIFF
--- a/utils/goutils.go
+++ b/utils/goutils.go
@@ -94,16 +94,23 @@ func getListCmdArgs() (cmdArgs []string, err error) {
 	return []string{"list", "-mod=mod"}, nil
 }
 
+type HandleErrorFunc func(err error) (bool, error)
+
 // Runs go list -f {{with .Module}}{{.Path}}:{{.Version}}{{end}} all command and returns map of the dependencies
-func GetDependenciesList(projectDir string, log Log) (map[string]bool, error) {
+func GetDependenciesList(projectDir string, log Log, handleError HandleErrorFunc) (map[string]bool, error) {
 	cmdArgs, err := getListCmdArgs()
 	if err != nil {
 		return nil, err
 	}
 	output, err := runDependenciesCmd(projectDir, append(cmdArgs, "-f", "{{with .Module}}{{.Path}}:{{.Version}}{{end}}", "all"), log)
 	if err != nil {
-		// Errors occurred while running "go list". Run again and this time ignore errors (with '-e')
 		log.Warn("Errors occurred while building the Go dependency tree. The dependency tree may be incomplete: " + err.Error())
+		if handleError != nil {
+			if stop, newErr := handleError(err); stop {
+				return nil, newErr
+			}
+		}
+		// Errors occurred while running "go list". Run again and this time ignore errors (with '-e')
 		output, err = runDependenciesCmd(projectDir, append(cmdArgs, "-e", "-f", "{{with .Module}}{{.Path}}:{{.Version}}{{end}}", "all"), log)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


utils/goutils.go
```diff
@@ -91,16 +91,23 @@ func getListCmdArgs() (cmdArgs []string, err error) {
 	return []string{"list", "-mod=mod"}, nil
 }
 
+type HandleErrorFunc func(err error) (bool, error)
```
attiasas, line 94: add comment on this function (usage and output variables)

utils/goutils.go
```diff
@@ -91,16 +91,23 @@ func getListCmdArgs() (cmdArgs []string, err error) {
...
 	if err != nil {
-		// Errors occurred while running "go list". Run again and this time ignore errors (with '-e')
 		log.Warn("Errors occurred while building the Go dependency tree. The dependency tree may be incomplete: " + err.Error())
+		if handleError != nil {
+			if stop, newErr := handleError(err); stop {
+				return nil, newErr
+			}
+		}
+		// Errors occurred while running "go list". Run again and this time ignore errors (with '-e')
 		output, err = runDependenciesCmd(projectDir, append(cmdArgs, "-e", "-f", "{{with .Module}}{{.Path}}:{{.Version}}{{end}}", "all"), log)
```
attiasas, line 111: what about the errors from "go list", why are we only handling the first cmd call?

attiasas (changes requested):
* Update PR description
* Update branch to latest dev.
* Please change the name of the PR to be more general. like "Handle specific errors when getting Go dependencies"
* Please add tests to this new cases

utils/goutils.go
asafambar: the second run ignores errors as it runs with -e, in general its the same command, so if it fails on 403 only on the second one(which cant happen as it ignores errors) it doesn't make sense

asafambar:
The PR introduces support for recognizing errors that occur while building the Go dependency tree. The primary reason for adding this support is to detect errors originating from the curation service. It's essential to catch these errors during the tree-building process, as they can occur when fetching dependencies from Artifactory as a result of running the go list command. The code changes involve passing a function to handle these errors, allowing the libraries that use it to manage error handling and reduce code duplication.

attiasas (approved):

utils/goutils_test.go
```diff
@@ -82,16 +82,24 @@ func TestGetProjectRoot(t *testing.T) {
...
+	testGetDependenciesList(t, "testGoList", nil)
 }
 
 func TestGetDependenciesListWithIgnoreErrors(t *testing.T) {
 	// In some cases, we see that running go list on some Go packages may fail.
 	// We should allow ignoring the errors in such cases and build the Go dependency tree, even if partial.
-	testGetDependenciesList(t, "testBadGoList")
+	testGetDependenciesList(t, "testBadGoList", nil)
+	// In some cases we would like to exit after we receive an error, this can be done with custom error handle func.
+	// this test handleErrorFunc return an error
```
eyalbe4, line 93: // In some cases we would like to exit after we receive an error, this can be done with custom error handle func.
// this test handleErrorFunc return an error
-->
// In some cases we would like to exit after we receive an error. This can be done with custom error handle func.
// This test handleErrorFunc return an error.

(We should avoid forcing two sentencing into one).

eyalbe4 (changes requested):

PR #273, opened 2024-09-05
Fix npm's deployment path

connyay:
Fixes https://github.com/jfrog/jfrog-cli/issues/2280

-----


connyay:

sverdlov93:
Hi @connyay, following the above fix do you think that the artifact name should include the scope also?
image

sverdlov93:
> Hi @connyay, following the above fix do you think that the artifact name should also include the scope? image because currently its uploaded without it . jfrog-cli-tests-=1.0.0.tgz > @scope/jfrog-cli-tests-=1.0.0.tgz

@connyay

connyay:
@sverdlov93 I am not sure. I don't think so. 

This test case matches my expectations:
https://github.com/jfrog/build-info-go/blob/c694d21943a912be849aada274dc0cc82cabdabc/build/utils/npm_test.go#L102

From what I can tell this change will fix https://github.com/jfrog/jfrog-cli/issues/2280

Is there a separate issue about the artifact name not having scope?

sverdlov93:
Following your change, we have failing tests on JFrog CLI, so I am trying to understand how to handle that.
For example:
{
  "name": "@my-scope/my-package",
  "version": "1.0.0",
  "description": "A sample scoped package",
}
The artifact name here should be @my-scope/my-package or my-package ?

PR #277, opened 2024-09-25
Skip install in dependencies map calculation if requested

eranturgeman:
-----

This PR contains changes to NPM installation process when calculating dependencies map.
We currently add a new ability to **skip installation** of a project if requested by the user.
This new functionality will be available through Frogbot and will currently apply only to Yarn and NPM (for WalkMe POC)
The changes reflected here are only in NPM files due to the way NPM dependencies map calculation is performed. In the near future we opt to re-write this logic and take the installation logic to cli-security and leave only the map construction in build-info-go.
Tests were added as well.
The new logic is as follows:
1) If the user has provided an 'install command' - we execute this command whether the project is installed or not (this is the logic we aim in all other PMs as well)
2) If there is no 'install command' and the project is not installed, while the user prohibited installation - we throw a new error to indicate that and handle this error later in the process
3) If there is no 'install command' and the project is not installed and the is no prohibition to install- we install using a default install command, as we do today

### DIFF
--- a/build/testdata/npm/noBuildProject/package.json
+++ b/build/testdata/npm/noBuildProject/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "npm_test2",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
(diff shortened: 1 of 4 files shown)


eyalbe4 (changes requested):
Let's create PRs for the jfrog-cli-core and jfrog-cli project, to make sure the tests there pass following this change.

PR #282, opened 2024-11-07
Apply allow-partial-results on Yarn V1 dependencies map construction

eranturgeman:
-----
This PR introduces an improvement that applies 'allow-partial-results' ability on Yarn V1 dependencies map construction, so if we have some missing dependencies - it will not fail the entire map construction.
**In more details:**
Due to the output of 'yarn list' we manually build the dependencies map for Yarn V1 in 2 phases. 
At the output's first level we get all dependencies (direct and indirect) with their locked versions.
At the output's second level (the children) we get all dependencies with caret version specifier (^).
Therefore a manual mapping between the versions is required for the final construction of the map and later the tree.
Sometime (for unknown reason that will be investigated further later, extra debug logs were added) we encounter a child dependency that we do now have at the first level of the map, and this causes the entire process to fail.
The current change states that if 'allow-partial-results' is enabled, and we encounter such a dependency - we just log the error and skip the matching for this problematic dependency.
This allow us to continue the process and to report the vulnerability, with a price that we might have some dependencies without impact paths (they will be reported in 'audit', just without stating their direct dependency) and will not get fixed in Frogbot.
**This applies ONLY if 'partial results' is enabled. In any other case the flow remains unchanged**

related changes: 
https://github.com/jfrog/jfrog-cli-security/pull/229
https://github.com/jfrog/frogbot/pull/784

### DIFF
--- a/build/utils/yarn.go
+++ b/build/utils/yarn.go
@@ -124,7 +124,6 @@ func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInf
 	if isV2AndAbove {
 		dependenciesMap, root, err = buildYarnV2DependencyMap(packageInfo, responseStr)
 	} else {
-		responseStr = "{\"type\":\"tree\",\"data\":{\"type\":\"list\",\"trees\":[{\"name\":\"minimist@1.2.5\",\"children\":[],\"hint\":null,\"color\":\"bold\",\"depth\":0},{\"name\":\"yarn-inner@1.0.0\",\"children\":[{\"name\":\"tough-cookie@2.5.0\",\"color\":\"dim\",\"shadow\":true}],\"hint\":null,\"color\":\"bold\",\"depth\":0}]}}"
 		dependenciesMap, root, err = buildYarnV1DependencyMap(packageInfo, responseStr, allowPartialResults, log)
 	}
 	return
--- a/build/utils/yarn.go
+++ b/build/utils/yarn.go
@@ -94,11 +94,11 @@ func GetYarnExecutable() (string, error) {
 	return yarnExecPath, nil
 }
 
-// GetYarnDependencies returns a map of the dependencies of a Yarn project and the root package of the project.
+// Returns a map of the dependencies of a Yarn project and the root package of the project.
 // The keys are the packages' values (Yarn's full identifiers of the packages), for example: '@scope/package-name@1.0.0'
 // (for yarn v < 2.0.0) or @scope/package-name@npm:1.0.0 (for yarn v >= 2.0.0).
 // Pay attention that a package's value won't necessarily contain its version. Use the version in package's details instead.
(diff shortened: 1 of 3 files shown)


build/utils/yarn.go
```diff
@@ -94,11 +94,11 @@ func GetYarnExecutable() (string, error) {
 	return yarnExecPath, nil
 }
 
-// GetYarnDependencies returns a map of the dependencies of a Yarn project and the root package of the project.
+// Returns a map of the dependencies of a Yarn project and the root package of the project.
 // The keys are the packages' values (Yarn's full identifiers of the packages), for example: '@scope/package-name@1.0.0'
 // (for yarn v < 2.0.0) or @scope/package-name@npm:1.0.0 (for yarn v >= 2.0.0).
 // Pay attention that a package's value won't necessarily contain its version. Use the version in package's details instead.
-func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInfo, log utils.Log) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
+func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInfo, log utils.Log, allowPartialResults bool) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
```
eyalbe4, line 101: Let's add to the function's description, the description for the function's arguments, and specifically the new `allowPartialResults` argument.
The same goes for the `buildYarnV1DependencyMap` function below.

build/utils/yarn.go
```diff
@@ -164,23 +164,27 @@ func buildYarnV1DependencyMap(packageInfo *PackageInfo, responseStr string) (dep
...
-
+		// We insert to dependenciesMap dependencies with the resolved versions only. All dependencies at the responseStr first level contain resolved versions only (their children may contain caret version ranges).
 		dependenciesMap[curDependency.Name] = &YarnDependency{
 			Value:   curDependency.Name,
 			Details: YarnDepDetails{Version: packageVersion},
 		}
 		packNameToFullName[packageCleanName] = curDependency.Name
 	}
+	log.Debug(fmt.Sprintf("package name to full name map content: %v", packNameToFullName))
+	log.Debug(fmt.Sprintf("'yarn list' output string: %s", responseStr))
```
eyalbe4, line 183: Let's merge the above two log entries into a single log entry.

build/utils/yarn.go
```diff
@@ -233,7 +242,7 @@ func buildYarnV2DependencyMap(packageInfo *PackageInfo, responseStr string) (dep
 	return
 }
 
-// runYarnInfoOrList depends on the yarn version currently operating on the project, runs the command that gets the dependencies of the project
+// Depends on the yarn version currently operating on the project, runs the command that gets the dependencies of the project
```
eyalbe4, line 245: I recommend a grammar checker, to make sure the comments have correct grammar.
Let's fix the above comment.

eyalbe4 (changes requested):

PR #301, opened 2025-05-06
Use Python3 Exe if exists for pip dependencies

attiasas:
-----

Add option to get pip dependencies using `Python3` and not only `python`

### DIFF
--- a/utils/pythonutils/piputils.go
+++ b/utils/pythonutils/piputils.go
@@ -19,7 +19,16 @@ func getPipDependencies(srcPath, dependenciesDirName string) (map[string][]strin
 	if err != nil {
 		return nil, nil, err
 	}
-	localPipdeptree := io.NewCommand("python", "", []string{localPipdeptreeScript, "--json"})
+	// Get the executable path of the python interpreter
+	var args []string
+	pythonExecutable, windowsPyArg := GetPython3Executable()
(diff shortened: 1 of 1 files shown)


utils/pythonutils/piputils.go
```diff
@@ -19,7 +19,16 @@ func getPipDependencies(srcPath, dependenciesDirName string) (map[string][]strin
 	if err != nil {
 		return nil, nil, err
 	}
-	localPipdeptree := io.NewCommand("python", "", []string{localPipdeptreeScript, "--json"})
+	// Get the executable path of the python interpreter
+	var args []string
+	pythonExecutable, windowsPyArg := GetPython3Executable()
```
EyalDelarea, line 24: Don't know if this is documented inside the function itself, but why does it has windows specific logic ?

EyalDelarea (changes requested):
the title isn't clear by the changes. 
How do you know based on your changes that you are in pip and you should use python3 ?

If the function that was already existed handles it, maybe we should change it's name :)

utils/pythonutils/piputils.go
attiasas: used a method that existed before.
It get you the python exe to run command (3 if supported, py for fallback)

EyalDelarea (approved):

PR #312, opened 2025-07-29
removed additional properties

reshmifrog:
-----

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -506,7 +506,6 @@ func createModule(moduleId string, moduleType entities.ModuleType, checksum enti
 func createDefaultModule(moduleId string) *entities.Module {
 	return &entities.Module{
 		Id:           moduleId,
-		Properties:   map[string][]string{},
 		Artifacts:    []entities.Artifact{},
 		Dependencies: []entities.Dependency{},
 	}


build/build.go
```diff
@@ -506,7 +506,6 @@ func createModule(moduleId string, moduleType entities.ModuleType, checksum enti
 func createDefaultModule(moduleId string) *entities.Module {
 	return &entities.Module{
 		Id:           moduleId,
```
fluxxBot, line 508: have we made sure this field is not getting utilised anywhere else?
even in the artifactory server code as well?
I can recall some discussion where this was being used for some docker related task.

fluxxBot (changes requested):

build/build.go
reshmifrog: I have verified that this properties field is not being used anywhere , for docker we are explicitly setting the properties field not using the one that is there in the `createDefaultModule`, hence it is safe to remove this field.

fluxxBot (approved):

build/build.go
bhanurp: @reshmifrog can you update this as dependency in jfrog-cli/jfrog-client-go/jfrog-cli-core/jfrog-cli-artifactory and make sure all tests are passing and also create npm build info and check how it is being used in artifactory-service

PR #323, opened 2025-09-15
Fix python empty dependency

bhanurp:
-----
## Problem Statement

The dependency collection for python package manager pypi is adding dependency Id as empty to differentiate between downloaded dependency file names and cached dependencies

## Solution

Add package name to dependency id and also add the scope of such dependency as cached to make sure build info stores it as cached dependency instead of downloaded

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -471,7 +471,7 @@ func addDependencyToPartialModule(dependency entities.Dependency, moduleId strin
 
 	if dependency.Id == "" || strings.TrimSpace(dependency.Id) == "" {
 		// Log this as it indicates a bug in dependency collection from cache
-		log.Debugf("Skipping dependency with empty ID. This may indicate a cache issue. Dependency SHA256: %s ModuleId: %s", dependency.Sha256, moduleId)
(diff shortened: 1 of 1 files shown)


build/build.go
```diff
@@ -467,6 +468,13 @@ func addArtifactToPartialModule(artifact entities.Artifact, moduleId string, par
 }
 
 func addDependencyToPartialModule(dependency entities.Dependency, moduleId string, partialModules map[string]*partialModule) {
+
+	if dependency.Id == "" || strings.TrimSpace(dependency.Id) == "" {
+		// Log this as it indicates a bug in dependency collection from cache
+		log.Debugf("Skipping dependency with empty ID. This may indicate a cache issue. Dependency SHA256: %s ModuleId: %s", dependency.Sha256, moduleId)
```
itsmeleela, line 474: I think log messages should avoid vague sentences like "This may" 

May be 

"Skipping dependency - empty ID detected (possible cache issue). Dependency SHA256: %s ModuleId: %s"

naveenku-jfrog (approved):

fluxxBot (approved):

PR #330, opened 2025-10-30
fix for git dependency parse error

nitinp19:
Problem :-
- when parsing the dependency other than the public dependency we are ending up with the parse error. In this case the dependency was from private git repository. 
- The parse error is coming mainly because the version field is missing from the dependency which leads to the error.

Solution :-
- checking for resolved field in the dependency and taking out the version using the commit


build/utils/npm.go
```diff
@@ -309,6 +312,39 @@ func (nld *npmLsDependency) id() string {
...
+
+func extractVersionFromGitUrl(gitUrl string) string {
+	parts := strings.Split(gitUrl, "#")
+	if len(parts) > 1 {
+		return parts[1]
+	}
+	return ""
+}
+
+// extracts a Git URL from npm problems array
```
bhanurp: add an example of for a given input what will be output in comments

nitinp19:
> just to confirm `parseDependencies` is not used in construction of build-info?

No, this function is used for the construction of build-info too

PR #376, opened 2026-04-15
RTECO-992 - Added support for uv package manager

agrasth:
---

Adds UV package manager support via the FlexPack pattern.

**New: `UVFlexPack`** — reads `uv.lock` (TOML) to collect dependency build-info without running any subprocess. Implements both `FlexPackManager` and `BuildInfoCollector` interfaces.

**Key behaviors:**
- Dependency IDs in `name:version` format; `sha256` from lock file hashes
- `requestedBy` chains built via BFS over the resolved dependency graph
- Dev dependencies excluded by default (`IncludeDevDependencies: false`); `InstalledPackages` map (from `uv pip list`) used as ground truth when available — correctly handles all `--no-dev`/`--only-dev`/`--group`/`--only-group` flag combinations without flag parsing
- Direct URL deps (`source = { url = "..." }`) and git deps (`source = { git = "..." }`) tracked with `DirectURL` field; `type` set to `"git"` for git sources; `GetDirectURLDeps()` exposes them so callers can skip Artifactory AQL enrichment for non-registry packages
- `LockFilePath` / `ProjectName` / `ProjectVersion` config overrides enable use with PEP 723 inline scripts (adjacent `script.py.lock` file, no `pyproject.toml`)
- `parseDependencies` split into focused helpers; `ensureParsed()` guard for idempotency; PEP 503 name normalization

**`mergeArtifacts` fix** (`entities/buildinfo.go`): when two artifacts share the same SHA1, prefer the one with a real Artifactory path over `path="."` — prevents build-step artifacts from blocking the correct path set by the publish step.

### DIFF
--- a/flexpack/uv_flexpack.go
+++ b/flexpack/uv_flexpack.go
@@ -13,26 +13,26 @@ import (
 	"github.com/jfrog/gofrog/log"
 )
 
-// UvLockFile represents the top-level structure of uv.lock
-type UvLockFile struct {
+// UVLockFile represents the top-level structure of uv.lock
+type UVLockFile struct {
 	Version  int          `toml:"version"`
 	Revision int          `toml:"revision"`
-	Packages []UvPackage  `toml:"package"`
+	Packages []UVPackage  `toml:"package"`
 }
 
-// UvPackage represents a [[package]] entry in uv.lock
-type UvPackage struct {
-	Name            string                        `toml:"name"`
-	Version         string                        `toml:"version"`
-	Source          UvSource                      `toml:"source"`
-	Dependencies    []UvDependencyEdge            `toml:"dependencies"`
-	DevDependencies map[string][]UvDependencyEdge `toml:"dev-dependencies"`
-	Sdist           *UvArtifact                   `toml:"sdist"`
-	Wheels          []UvArtifact                  `toml:"wheels"`
+// UVPackage represents a [[package]] entry in uv.lock
+type UVPackage struct {
+	Name            string                         `toml:"name"`
+	Version         string                         `toml:"version"`
+	Source          UVSource                       `toml:"source"`
+	Dependencies    []UVDependencyEdge             `toml:"dependencies"`
+	DevDependencies map[string][]UVDependencyEdge  `toml:"dev-dependencies"`
+	Sdist           *UVArtifact                    `toml:"sdist"`
+	Wheels          []UVArtifact                   `toml:"wheels"`
 }
 
-// UvSource is an inline table with exactly one key identifying the source type.
-type UvSource struct {
+// UVSource is an inline table with exactly one key identifying the source type.
+type UVSource struct {
 	Registry  string `toml:"registry"`
 	Virtual   string `toml:"virtual"`
 	Editable  string `toml:"editable"`
@@ -42,12 +42,12 @@ type UvSource struct {
 }
 
 // IsWorkspacePackage returns true if this source represents a local workspace package.
-func (s UvSource) IsWorkspacePackage() bool {
+func (s UVSource) IsWorkspacePackage() bool {
 	return s.Virtual != "" || s.Editable != "" || s.Directory != ""
 }
 
-// UvArtifact represents an sdist or wheel entry in uv.lock
-type UvArtifact struct {
+// UVArtifact represents an sdist or wheel entry in uv.lock
+type UVArtifact struct {
 	URL        string `toml:"url"`
 	Path       string `toml:"path"`
 	Hash       string `toml:"hash"`        // "sha256:<hex>"; absent for git
@@ -55,27 +55,27 @@ type UvArtifact struct {
 	UploadTime string `toml:"upload-time"` // ISO 8601; may be absent (revision < 3)
 }
 
-// UvDependencyEdge represents a dependency reference inside a [[package]] entry
-type UvDependencyEdge struct {
+// UVDependencyEdge represents a dependency reference inside a [[package]] entry
+type UVDependencyEdge struct {
 	Name    string   `toml:"name"`
 	Marker  string   `toml:"marker"`
 	Extra   []string `toml:"extra"`
 	Version string   `toml:"version"`
 }
 
-// UvPyProjectToml reads only [project] (PEP 621) — UV format, not Poetry
-type UvPyProjectToml struct {
+// UVPyProjectToml reads only [project] (PEP 621) — UV format, not Poetry
+type UVPyProjectToml struct {
 	Project struct {
 		Name    string `toml:"name"`
 		Version string `toml:"version"`
 	} `toml:"project"`
 }
 
-// UvFlexPack implements FlexPackManager and BuildInfoCollector for the UV package manager
-type UvFlexPack struct {
-	config            UvConfig
-	lockFileData      *UvLockFile
-	pyprojectData     *UvPyProjectToml
+// UVFlexPack implements FlexPackManager and BuildInfoCollector for the UV package manager
+type UVFlexPack struct {
+	config            UVConfig
+	lockFileData      *UVLockFile
+	pyprojectData     *UVPyProjectToml
 	projectName       string
 	projectVersion    string
 	parsed            bool
@@ -84,9 +84,9 @@ type UvFlexPack struct {
 	requestedByChains map[string][][]string // dep ID -> full chains back to root (UV-specific)
 }
 
-// NewUvFlexPack creates a new UvFlexPack instance.
-func NewUvFlexPack(config UvConfig) (*UvFlexPack, error) {
-	uf := &UvFlexPack{
+// NewUVFlexPack creates a new UVFlexPack instance.
+func NewUVFlexPack(config UVConfig) (*UVFlexPack, error) {
+	uf := &UVFlexPack{
 		config:            config,
 		dependencies:      []DependencyInfo{},
 		depGraph:          make(map[string][]string),
@@ -102,13 +102,13 @@ func NewUvFlexPack(config UvConfig) (*UvFlexPack, error) {
 }
 
 // loadPyProjectToml reads and parses pyproject.toml using PEP 621 [project] section.
-func (uf *UvFlexPack) loadPyProjectToml() error {
+func (uf *UVFlexPack) loadPyProjectToml() error {
 	pyprojectPath := filepath.Join(uf.config.WorkingDirectory, "pyproject.toml")
 	data, err := os.ReadFile(pyprojectPath)
 	if err != nil {
 		return fmt.Errorf("failed to read pyproject.toml: %w", err)
 	}
-	uf.pyprojectData = &UvPyProjectToml{}
+	uf.pyprojectData = &UVPyProjectToml{}
 	if err := toml.Unmarshal(data, uf.pyprojectData); err != nil {
 		return fmt.Errorf("failed to parse pyproject.toml: %w", err)
 	}
@@ -124,13 +124,13 @@ func (uf *UvFlexPack) loadPyProjectToml() error {
 }
 
 // loadUvLock reads and parses uv.lock.
-func (uf *UvFlexPack) loadUvLock() error {
+func (uf *UVFlexPack) loadUvLock() error {
 	lockPath := filepath.Join(uf.config.WorkingDirectory, "uv.lock")
 	data, err := os.ReadFile(lockPath)
 	if err != nil {
 		return fmt.Errorf("failed to read uv.lock: %w", err)
 	}
-	uf.lockFileData = &UvLockFile{}
+	uf.lockFileData = &UVLockFile{}
 	if err := toml.Unmarshal(data, uf.lockFileData); err != nil {
 		return fmt.Errorf("failed to parse uv.lock: %w", err)
 	}
@@ -151,7 +151,7 @@ func extractSHA256(hash string) string {
 
 // bestHash returns the best available hash for a package.
 // Prefers pure-Python wheel (none-any), falls back to first wheel, then sdist.
-func bestHash(pkg UvPackage) string {
+func bestHash(pkg UVPackage) string {
 	for _, w := range pkg.Wheels {
 		if strings.Contains(w.URL, "none-any") && w.Hash != "" {
 			return w.Hash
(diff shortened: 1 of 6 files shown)


flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+// parseDependencies populates uf.dependencies and uf.depGraph from the lock file.
+// ID format and requestedBy chains match the pip/pipenv canonical build-info format:
+//   - dep ID:  "name:version"  (e.g. "certifi:2026.2.25")
+//   - dep type: file extension (e.g. "whl", "tar.gz")
+//   - requestedBy: full chain back to root module (e.g. [["requests:2.33.1","myapp:0.1.0"]])
+//   - direct deps: requestedBy = [["myapp:0.1.0"]]
+//   - no scopes (Python has no compile/runtime distinction; matches pip/pipenv)
+func (uf *UvFlexPack) parseDependencies() {
```
bhanurp: Please refactor this function

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,577 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/jfrog/build-info-go/entities"
+	"github.com/jfrog/gofrog/log"
+)
+
+// UvLockFile represents the top-level structure of uv.lock
+type UvLockFile struct {
```
bhanurp: Identifiers like uv should be all uppercase or if it is private struct or func which starts with uv should be all lowercase. Don't mix. Please make the changes across all repositories where ever applied to UV uppercase

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,577 @@
...
+	Project struct {
+		Name    string `toml:"name"`
+		Version string `toml:"version"`
+	} `toml:"project"`
+}
+
+// UvFlexPack implements FlexPackManager and BuildInfoCollector for the UV package manager
+type UvFlexPack struct {
+	config            UvConfig
+	lockFileData      *UvLockFile
```
bhanurp: Is lock file mandatory to get list of all dependencies? Isn't there any uv command which can be used to fetch list of dependencies?

bhanurp (changes requested):

flexpack/uv_flexpack.go
agrasth: Done — `parseDependencies` has been broken into six focused helpers: `buildPackageMap`, `findRootPackage`, `collectDirectDeps`, `buildDepInfoMap`, `buildForwardGraph`, and `collectRootChildren`. The orchestrating function is now ~25 lines.

flexpack/uv_flexpack.go
agrasth: Fixed — all exported identifiers now use the `UV` prefix (`UVFlexPack`, `UVConfig`, `UVLockFile`, etc.) per Go's convention for 2-letter initialisms. Unexported helpers retain the lowercase `uv` prefix. Changes applied across all four repos.

flexpack/uv_flexpack.go
agrasth: The lock file is the right source for build info for several reasons: 
(1) `uv.lock` is always present after any `uv sync`/`uv lock` run and contains the full resolved dependency tree with SHA-256 checksums — no additional network calls needed. 
(2) `uv tree` outputs human-readable text that is not reliably machine-parseable. 
(3) `uv export` produces a flat requirements list without the tree structure needed for `requestedBy` chains. 
(4) This is consistent with how Poetry (`poetry.lock`) and pip (requirements.txt hashes) are handled in this codebase. The lock file is the single source of truth for the resolved graph.

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+	}
+	return uf, nil
+}
+
+// loadPyProjectToml reads and parses pyproject.toml using PEP 621 [project] section.
+func (uf *UVFlexPack) loadPyProjectToml() error {
+	// When ProjectName is supplied via config (e.g. for PEP 723 inline scripts that
+	// have no pyproject.toml), skip file loading and use the overrides directly.
+	if uf.config.ProjectName != "" {
+		uf.projectName = uf.config.ProjectName
```
naveenku-jfrog, line 109: What if uf.config.ProjectVersion is nil?

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+}
+
+// extractSHA256 strips the "sha256:" prefix from a hash string.
+func extractSHA256(hash string) string {
+	return strings.TrimPrefix(hash, "sha256:")
+}
+
+// bestHash returns the best available hash for a package.
+// Prefers pure-Python wheel (none-any), falls back to first wheel, then sdist.
+func bestHash(pkg UVPackage) string {
```
naveenku-jfrog: can we use some more meaningful name instead of bestHash?

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+		return pkg.Sdist.Hash
+	}
+	return ""
+}
+
+// depFileType returns the file extension type for a package artifact.
+// Prefers pure-Python wheel (none-any), falls back to first wheel, then sdist.
+// Matches pip/pipenv behavior: "whl", "tar.gz", "zip", etc.
+func depFileType(pkg UVPackage) string {
+	bestURL := ""
```
naveenku-jfrog: can we use some more meaningful name instead of bestURL?

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+func normalizeName(name string) string {
+	return pep503Re.ReplaceAllString(strings.ToLower(name), "-")
+}
+
+// extractSHA256 strips the "sha256:" prefix from a hash string.
+func extractSHA256(hash string) string {
+	return strings.TrimPrefix(hash, "sha256:")
+}
+
+// bestHash returns the best available hash for a package.
```
naveenku-jfrog:
```suggestion
// bestHash returns the best available hash for a package.
func bestHash(pkg UVPackage) string {
	var firstWheelHash string

	for _, w := range pkg.Wheels {
		if w.Hash == "" {
			continue
		}
		
		// If it's the gold standard (universal), return immediately
		if strings.Contains(w.URL, "none-any") {
			return w.Hash
		}

		// Otherwise, save the first wheel hash we see as a fallback
		if firstWheelHash == "" {
			firstWheelHash = w.Hash
		}
	}

	// If we found any wheel at all, return it
	if firstWheelHash != "" {
		return firstWheelHash
	}

	// Final fallback to source distribution
	if pkg.Sdist != nil && pkg.Sdist.Hash != "" {
		return pkg.Sdist.Hash
	}

	return ""
}
```

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+	}
+	base := filepath.Base(bestURL)
+	if strings.HasSuffix(base, ".tar.gz") {
+		return "tar.gz"
+	}
+	if i := strings.LastIndex(base, "."); i != -1 {
+		return base[i+1:]
+	}
+	return ""
+}
```
naveenku-jfrog, line 222:
```suggestion
}
func depFileType(pkg UVPackage) string {
	var bestURL string

	// 1. Find the best available URL
	for _, w := range pkg.Wheels {
		if w.URL == "" {
			continue
		}
		if strings.Contains(w.URL, "none-any") {
			bestURL = w.URL
			break // Perfection found
		}
		if bestURL == "" {
			bestURL = w.URL // Keep the first specific wheel as a backup
		}
	}

	if bestURL == "" && pkg.Sdist != nil {
		bestURL = pkg.Sdist.URL
	}

	// 2. Handle Git if no URL was found
	if bestURL == "" {
		if pkg.Source.Git != "" {
			return "git"
		}
		return ""
	}

	// 3. Extract Format
	base := filepath.Base(bestURL)
	if strings.HasSuffix(base, ".tar.gz") {
		return "tar.gz"
	}
	// Using Ext is slightly cleaner than LastIndex for standard cases
	ext := filepath.Ext(base)
	if ext != "" {
		return strings.TrimPrefix(ext, ".")
	}

	return ""
}
```

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+		directURL := pkg.Source.URL
+		if directURL == "" {
+			directURL = pkg.Source.Git
+		}
+		depInfoMap[normalizedName] = &DependencyInfo{
+			ID:        fmt.Sprintf("%s:%s", pkg.Name, pkg.Version),
+			Name:      pkg.Name,
+			Version:   pkg.Version,
+			Type:      depFileType(*pkg),
+			SHA256:    extractSHA256(bestHash(*pkg)),
```
naveenku-jfrog: We are not considering sha1, md5 here?

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+		idGraph[depInfoMap[normalizedName].ID] = func() []string {
+			var ids []string
+			for _, c := range children {
+				ids = append(ids, depInfoMap[c].ID)
+			}
+			return ids
+		}()
+	}
+	return fwdGraph
+}
```
naveenku-jfrog, line 385:
```suggestion
}
func buildForwardGraph(depInfoMap map[string]*DependencyInfo, pkgByName map[string]*UVPackage, idGraph map[string][]string) map[string][]string {
	fwdGraph := make(map[string][]string, len(depInfoMap))
	
	for normalizedName, info := range depInfoMap {
		pkg := pkgByName[normalizedName]
		if pkg == nil {
			continue
		}

		// Pre-allocate based on dependency count to reduce re-allocations
		children := make([]string, 0, len(pkg.Dependencies))
		for _, edge := range pkg.Dependencies {
			childName := normalizeName(edge.Name)
			// Ensure the child is actually part of our filtered dependency set
			if _, ok := depInfoMap[childName]; ok {
				children = append(children, childName)
			}
		}
		fwdGraph[normalizedName] = children

		// Populate ID graph without anonymous function overhead
		childIDs := make([]string, len(children))
		for i, c := range children {
			childIDs[i] = depInfoMap[c].ID
		}
		idGraph[info.ID] = childIDs
	}
	return fwdGraph
}
```

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+		}
+		if len(chains[child.ID]) >= maxDepth {
+			continue
+		}
+		// New chain entry: [parentID, ...parentChain]
+		newChain := append([]string{parentID}, parentChain...)
+		// Cycle check: if child's own ID already appears in the chain, skip
+		for _, id := range newChain {
+			if id == child.ID {
+				goto next
```
naveenku-jfrog: Use of goto statement is discouraged.

flexpack/uv_flexpack.go
```diff
@@ -0,0 +1,697 @@
...
+	for i := range packages {
+		pkg := &packages[i]
+		if pkg.Source.IsWorkspacePackage() {
+			continue
+		}
+		normalizedName := normalizeName(pkg.Name)
+		if !includeDevDeps {
+			if mainReachable != nil {
+				if !mainReachable[normalizedName] {
+					continue // reachability analysis: exclude dev-only transitive deps
```
naveenku-jfrog: Multiple if statements here, can we combine them and simplify?

naveenku-jfrog (changes requested):

flexpack/uv_flexpack.go
agrasth: By design — `uv.lock` only records sha256 hashes (from the wheel/sdist archive). sha1 and md5 are enriched separately in `jfrog-cli-artifactory` by querying Artifactory's AQL endpoint after sync. The FlexPack's responsibility is to read what the lock file provides; the calling layer handles Artifactory enrichment.

flexpack/uv_flexpack.go
agrasth: `ProjectVersion` is a `string` (not a pointer) so it cannot be `nil`, but it can be empty. Added a comment clarifying that an empty version is intentional for PEP 723 inline scripts — those have no `pyproject.toml` so no version is available. 
The module ID becomes `"scriptname:"` in that case, which is acceptable since the caller controls the build-info context.

flexpack/uv_flexpack.go
agrasth: Renamed to `selectPackageHash` — it better describes the intent: selecting the best available hash from a package's artifacts, preferring universal wheels over platform-specific ones and falling back to sdist.

flexpack/uv_flexpack.go
agrasth: Renamed the local variable to `selectedURL` and refactored `depFileType` to a single-pass loop with early `continue` on empty URLs — cleaner logic, same result.

flexpack/uv_flexpack.go
agrasth: Applied the suggestion. `selectPackageHash` (renamed from `bestHash`) now uses a single pass: skips empty hashes via `continue`, returns immediately for universal wheels (`none-any`), tracks the first platform wheel as fallback, then falls through to sdist.

flexpack/uv_flexpack.go
agrasth: Applied the suggestion. `depFileType` now uses a single-pass loop with early `continue` on empty URLs, preferring universal wheels and falling back to the first platform wheel, then sdist. The local variable is renamed to `selectedURL` for clarity.

flexpack/uv_flexpack.go
agrasth: Applied. `buildForwardGraph` now iterates with `for normalizedName, info := range depInfoMap` and uses `info.ID` directly, eliminating the redundant map lookup.

flexpack/uv_flexpack.go
agrasth: Removed. Replaced `goto next` / `next:` with a `hasCycle bool` flag and a guard `if !hasCycle { ... }` block — cleaner and no label needed.

flexpack/uv_flexpack.go
agrasth: Simplified. The nested ifs are now a single flattened boolean: `excluded := (mainReachable != nil && !mainReachable[normalizedName]) || (mainReachable == nil && directDevDeps[normalizedName])` followed by `if excluded { continue }`.

naveenku-jfrog (approved):

PR #378, opened 2026-05-07
added nix package manager support

reshmifrog:
-----
Added the support for nix package manager


Balasubramanyamkosuri (commented):
## Review summary

Thanks for the Nix flexpack work — the `flexpack/nix/` package and unit tests look solid.

### Blocker: `mergeArtifacts` regression

This PR removes logic in `entities/buildinfo.go` that prefers a non-`.` artifact path when SHA1 matches:

```go
if existingArtifact.Path == "." && newArtifact.Path != "." {
    (*intoArtifacts)[i] = newArtifact
}
```

That path preference is required for native UV/Poetry publish flows: artifacts are first recorded locally (`path="."`), then merged again after Artifactory upload with the real repo path. Without it, publish build-info can keep stale local paths and break property stamping / artifact browser links.

**Please revert that hunk or add regression tests** (UV publish ITs in jfrog-cli cover this).

### Questions

1. `bi nix` uses hard-coded `CollectBuildInfo("nix-build", "1")` — is that intentional for a follow-up?
2. Confirm PR targets the correct base branch per repo policy.

Full structured review: see team doc `docs/reviews/nix-pr-review.md` in the Nix workspace.

flexpack/nix/storepath.go
naveenku-jfrog, line 1: This file name is not allowed.

flexpack/nix/flexpack.go
naveenku-jfrog, line 1: This file name is not allowed.

flexpack/nix/nix_types.go
naveenku-jfrog, line 1: This file name is not allowed.

flexpack/nix/narinfo_parser.go
```diff
@@ -0,0 +1,112 @@
...
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		idx := strings.Index(line, ": ")
+		if idx == -1 {
+			continue
+		}
+		key := line[:idx]
+		value := line[idx+2:]
```
naveenku-jfrog: There is no check if idx+2 < length(line).

flexpack/nix/narinfo_parser.go
```diff
@@ -0,0 +1,112 @@
...
+	return base
+}
+
+// ExtractNameAndVersion splits a package name into name and version.
+// e.g. "hello-2.12.3" → ("hello", "2.12.3")
+// e.g. "glibc-2.40-66" → ("glibc", "2.40-66")
+// e.g. "bash" → ("bash", "")
+func ExtractNameAndVersion(packageName string) (string, string) {
+	for i := len(packageName) - 1; i >= 0; i-- {
+		if packageName[i] == '-' && i+1 < len(packageName) && packageName[i+1] >= '0' && packageName[i+1] <= '9' {
```
naveenku-jfrog: what is relevance of '9'?

cli/cli.go
```diff
@@ -290,6 +291,34 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
+					return fmt.Errorf("nix build-info collection requires FlexPack mode (set JFROG_RUN_NATIVE=true)")
+				}
+				config := nixflex.NixChannelConfig{
+					WorkingDirectory: ".",
+				}
+				collector, err := nixflex.NewNixChannelCollector(config)
+				if err != nil {
+					return fmt.Errorf("failed to create Nix collector: %w", err)
+				}
+				buildInfo, err := collector.CollectBuildInfo("nix-build", "1")
```
naveenku-jfrog, line 324: CollectStorePathDependencies is never called before this, so c.dependencies and c.dependencyGraph are always empty and every invocation of bi nix returns a build-info with one empty nix module. Please either accept store paths as positional args (bi nix <storePath>...) and call collector.CollectStorePathDependencies(args...) first, or auto-discover the build's roots (e.g. ./result, nix-store --query --references ...) before calling CollectBuildInfo. Without this, the command does nothing useful.

cli/cli.go
```diff
@@ -290,6 +291,34 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
+				if err != nil {
+					return fmt.Errorf("failed to collect build info: %w", err)
+				}
+				formatValue, _, err := extractStringFlag(context.Args().Slice(), formatFlag)
```
naveenku-jfrog: parse the format flag before calling CollectBuildInfo so a typo in --format doesn't make users wait for a nix path-info shell-out before failing.

entities/buildinfo.go
```diff
@@ -271,14 +272,9 @@ func mergeArtifacts(mergeArtifacts *[]Artifact, intoArtifacts *[]Artifact) {
 	for _, newArtifact := range *mergeArtifacts {
 		exists := false
 
-		// PRIORITY 1: Check SHA1 - if checksums match, prefer the artifact with a real path.
-		// path="." means the artifact was recorded locally before upload; a non-"." path
-		// means it was confirmed in Artifactory. Always keep the richer entry.
-		for i, existingArtifact := range *intoArtifacts {
+		// PRIORITY 1: Check SHA1 - if checksums match, artifact already exists (skip it)
+		for _, existingArtifact := range *intoArtifacts {
```
naveenku-jfrog: This change is unrelated to adding Nix support and should be split into its own PR.

flexpack/nix/nix_channel_collector.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		dep := flexpack.DependencyInfo{
+			ID:      depID,
+			Name:    name,
+			Version: version,
+			SHA256:  sha256,
+			Scopes:  []string{"runtime"},
+			Path:    path,
+		}
```
naveenku-jfrog: SHA256 is being filled from info.NarHash (the hash of the serialized NAR archive), but Artifactory will compute sha256 over the uploaded file (.nar.xz or the unpacked output). These two values will never match, so any downstream that correlates build-info deps to repo files by checksum (Xray scan linkage, dedupe, property stamping by sha256) will silently fail.

Either (a) populate SHA256 from whatever file is actually uploaded to Artifactory and surface narHash separately as a property, or (b) document explicitly that this field is the NAR hash and not a content checksum

naveenku-jfrog (changes requested):

entities/buildinfo.go
```diff
@@ -271,14 +272,9 @@ func mergeArtifacts(mergeArtifacts *[]Artifact, intoArtifacts *[]Artifact) {
...
 			if newArtifact.Sha1 == existingArtifact.Sha1 {
-				if existingArtifact.Path == "." && newArtifact.Path != "." {
```
agrasth, line 279: The simplification of mergeArtifacts removes the path-priority logic ("prefer artifact with real path over '.'"). This is a cross-cutting behavior change affecting all package managers, not just Nix.
Right now it looks like an accidental cleanup.

flexpack/nix/narinfo_parser.go
```diff
@@ -0,0 +1,112 @@
...
+	"encoding/base64"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// ParseNarInfo parses a .narinfo file's plain-text content into a NixNarInfo struct.
+// Format is line-based key-value: "Key: Value"
+func ParseNarInfo(content string) (*NixNarInfo, error) {
```
agrasth: ParseNarInfo has an error return but always returns nil. Either use it to signal malformed input, or change the signature to `func ParseNarInfo(content string) *NixNarInfo`. Callers shouldn't have to handle an error that can never occur.

flexpack/nix/narinfo_parser.go
```diff
@@ -0,0 +1,112 @@
...
+		case "StorePath":
+			info.StorePath = value
+		case "URL":
+			info.URL = value
+		case "Compression":
+			info.Compression = value
+		case "FileHash":
+			info.FileHash = value
+		case "FileSize":
+			info.FileSize, _ = strconv.ParseInt(value, 10, 64)
```
agrasth: `strconv.ParseInt` errors are silently discarded with _. 
A malformed FileSize or NarSize field will silently produce 0. Should we log a warning, WDYT?

flexpack/nix/nix_channel_collector.go
```diff
@@ -0,0 +1,230 @@
...
+		if rootIDs[depID] {
+			continue
+		}
+
+		pkgName := ExtractPackageName(path)
+		name, version := ExtractNameAndVersion(pkgName)
+
+		// Convert narHash from SRI to hex for Artifactory matching
+		sha256 := info.NarHash
+		if hexHash, err := SriToHex(info.NarHash); err == nil {
```
agrasth: If `SriToHex` fails (non-sha256 narHash, or unpadded base64), the raw NarHash string (e.g., `sha256:0y8xki...` in `Nix32` encoding) is stored as the checksum. Artifactory won't match this. Should log a warning and leave SHA256 empty rather than storing an unrecognizable value.

flexpack/nix/nix_channel_collector.go
```diff
@@ -0,0 +1,230 @@
...
+		entityDep := entities.Dependency{
+			Id:     dep.ID,
+			Scopes: dep.Scopes,
+			Checksum: entities.Checksum{
+				Sha256: dep.SHA256,
+			},
+		}
+
+		if parents, exists := requestedBy[dep.ID]; exists && len(parents) > 0 {
+			entityDep.RequestedBy = [][]string{parents}
```
agrasth: RequestedBy is set as [][]string{parents} where parents is []string{"parent1", "parent2"}. This produces [["parent1", "parent2"]] — one chain with two entries — rather than [["parent1"], ["parent2"]] (two separate request chains). 
You can take a look at how Go or npm modules populate RequestedBy to match the expected format.

flexpack/nix/nix_channel_collector.go
```diff
@@ -0,0 +1,230 @@
...
+
+		c.dependencies = append(c.dependencies, dep)
+	}
+
+	log.Debug(fmt.Sprintf("Collected %d runtime dependencies from store closure", len(c.dependencies)))
+	return nil
+}
+
+// CollectBuildInfo creates a complete BuildInfo from the collected dependencies.
+func (c *NixChannelCollector) CollectBuildInfo(buildName, buildNumber string) (*entities.BuildInfo, error) {
```
agrasth: Agent.Version: "1.0" is hardcoded. 
Possible to run nix --version and parse the output to populate this accurately? the same way other collectors get the tool version.

flexpack/nix/nix_types.go
```diff
@@ -0,0 +1,35 @@
...
+
+// NixChannelConfig holds configuration for Nix channel-based build-info collection.
+type NixChannelConfig struct {
+	WorkingDirectory string
+	ChannelName      string // e.g. "nixos-25.11"
+}
+
+// NixStorePathInfo represents a single entry from "nix path-info --json -r" output.
+// The JSON output is a map keyed by store path.
+type NixStorePathInfo struct {
```
agrasth: `NixNarInfo.References` is string (space-separated), while `NixStorePathInfo.References` is `[]string`. Is this expected?
The `NarInfo` struct is only used in unit tests — nothing in the collector actually calls ParseNarInfo. Either integrate it into the collector's workflow or note clearly that it's a utility for future use.

flexpack/nix/storepath.go
fluxxBot, line 1: let's rename this file to more appropriate descriptive name

flexpack/nix/nix_types.go
fluxxBot, line 1: put this in the respective files

fluxxBot (changes requested):

entities/buildinfo.go
```diff
@@ -271,14 +272,9 @@ func mergeArtifacts(mergeArtifacts *[]Artifact, intoArtifacts *[]Artifact) {
 	for _, newArtifact := range *mergeArtifacts {
 		exists := false
 
-		// PRIORITY 1: Check SHA1 - if checksums match, prefer the artifact with a real path.
-		// path="." means the artifact was recorded locally before upload; a non-"." path
-		// means it was confirmed in Artifactory. Always keep the richer entry.
-		for i, existingArtifact := range *intoArtifacts {
+		// PRIORITY 1: Check SHA1 - if checksums match, artifact already exists (skip it)
```
fluxxBot: why did we removed it?

flexpack/nix/narinfo_parser.go
fluxxBot: check if we can use this package instead of writing the parser ourself - https://github.com/nix-community/go-nix

flexpack/nix/storepath.go
```diff
@@ -0,0 +1,112 @@
...
+	if idx := strings.Index(base, "-"); idx != -1 {
+		return base[idx+1:]
+	}
+	return base
+}
+
+// ExtractNameAndVersion splits a package name into name and version.
+// e.g. "hello-2.12.3" → ("hello", "2.12.3")
+// e.g. "glibc-2.40-66" → ("glibc", "2.40-66")
+// e.g. "bash" → ("bash", "")
```
fluxxBot, line 38: why are we doing this? is this correct?

flexpack/nix/flexpack.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+// NixChannelCollector implements FlexPackManager and BuildInfoCollector for
+// Nix channel-based workflows. It collects dependencies from the Nix store
+// after a package has been installed/built via channels.
+type NixChannelCollector struct {
+	config          NixChannelConfig
+	dependencies    []flexpack.DependencyInfo
+	dependencyGraph map[string][]string
+	projectName     string
```
fluxxBot, line 42: why is it missing artifacts?

flexpack/nix/flexpack.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// CollectStorePathDependencies runs "nix path-info --json -r" on the given store paths
+// and collects the runtime closure as dependencies.
+func (c *NixChannelCollector) CollectStorePathDependencies(storePaths ...string) error {
+	if len(storePaths) == 0 {
+		return nil
+	}
+
+	args := append([]string{"path-info", "--json", "--recursive"}, storePaths...)
```
fluxxBot, line 95: try to use this go-package - https://github.com/nix-community/go-nix

fluxxBot, line 95: why are we using this then - https://github.com/jfrog/build-info-go/pull/378/changes#diff-7d1f29a8712a6ccff2f9208e569fa7ba9cb5e59dad12fa4661c48acaeb1594e6R28-R29?

flexpack/nix/flexpack.go
```diff
@@ -0,0 +1,230 @@
...
+	return c.dependencyGraph, nil
+}
+
+// GetDependency returns a formatted string summary of dependencies.
+func (c *NixChannelCollector) GetDependency() string {
+	var result strings.Builder
+	fmt.Fprintf(&result, "Project: %s\n", c.projectName)
+	result.WriteString("Dependencies:\n")
+	for _, dep := range c.dependencies {
+		fmt.Fprintf(&result, "  - %s [%s]\n", dep.ID, dep.SHA256)
```
fluxxBot, line 305: whare are we getting this sha256 from?

flexpack/nix/flexpack.go
fluxxBot, line 1: also why is this file named after channel, how will flakes work here?

fluxxBot (changes requested):

flexpack/nix/nix_flexpack.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+		BuildAgent: &entities.Agent{
+			Name:    "Nix",
+			Version: nixVersion,
+		},
+		Modules: []entities.Module{},
+	}
+
+	module := entities.Module{
+		Id:   c.projectName,
```
naveenku-jfrog: We are not using module flag value here?

naveenku-jfrog (changes requested):

naveenku-jfrog (approved):

PR #379, opened 2026-05-07
Enhance Yarn module to support workspace projects and non-registry de…

fluxxBot:
…pendencies. Added tests for workspace detection and locator protocols. Updated dependency retrieval to handle Yarn versioning more efficiently.

-----

RTECO-1035 — Fix yarn workspace and non-registry dependency collection                                                                                                              
                                                                                                                                                                                      
  Problem                                                                                                                                                                             
                                                                                                                                                                                      
  Three classes of build-info collection bugs affecting yarn v2+:                                                                                                                     
                                      
  1. Workspace monorepo root install produces empty build-info. yarn info --all --recursive --json emits all deps but the root locator (s1-root@workspace:.) has no Dependencies edges
   — the walker starts at root, finds nothing, terminates. All registry deps installed by member packages are unreachable.
  2. Non-registry locators emitted as ghost dependencies with misleading warning. Deps using link:, file:, portal:, git+https:, workspace:* protocols are local/VCS sources — they    
  don't exist in Artifactory. The walker attempted AQL lookup, failed, then printed "could not be found in Artifactory and therefore are not included in the build-info" for every    
  such dep. Warning is incorrect — these deps are intentionally excluded.
  3. yarn --version executed twice per build. newYarnModule calls validateYarnVersion → GetVersion, then getDependenciesMap → GetYarnDependencies → GetVersion again. Redundant       
  subprocess.

### DIFF
--- a/build/utils/yarn.go
+++ b/build/utils/yarn.go
@@ -105,12 +105,20 @@ func GetYarnExecutable() (string, error) {
 // log - The logger.
 // allowPartialResults - If true, the function will allow some errors to occur without failing the flow and will generate partial results.
 func GetYarnDependencies(executablePath, srcPath string, packageInfo *PackageInfo, log utils.Log, allowPartialResults bool) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
-	executableVersionStr, err := GetVersion(executablePath, srcPath)
-	if err != nil {
-		return
+	return GetYarnDependenciesWithVersion(executablePath, srcPath, packageInfo, "", log, allowPartialResults)
+}
+
+// GetYarnDependenciesWithVersion is the same as GetYarnDependencies but accepts a pre-fetched
+// yarn version string to avoid re-running 'yarn --version'. Pass an empty string to fetch it.
+func GetYarnDependenciesWithVersion(executablePath, srcPath string, packageInfo *PackageInfo, yarnVersion string, log utils.Log, allowPartialResults bool) (dependenciesMap map[string]*YarnDependency, root *YarnDependency, err error) {
+	if yarnVersion == "" {
+		yarnVersion, err = GetVersion(executablePath, srcPath)
+		if err != nil {
+			return
+		}
 	}
 
-	isV2AndAbove := version.NewVersion(executableVersionStr).Compare(yarnV2Version) <= 0
+	isV2AndAbove := version.NewVersion(yarnVersion).Compare(yarnV2Version) <= 0
 
 	// Run 'yarn info or list'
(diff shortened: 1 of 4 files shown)


build/utils/yarn.go
```diff
@@ -264,6 +272,30 @@ func buildYarnV2DependencyMap(packageInfo *PackageInfo, responseStr string) (dep
 }
 
 // Depending on the Yarn version currently in use for the project, this function runs the command that retrieves the project's dependencies
+// IsYarnWorkspaceProject returns true when the project at srcPath is a yarn workspace
+// (monorepo). It runs `yarn workspaces list --json` and checks whether the output
+// contains more than one entry — a workspace project always lists the root plus at
+// least one member, while a plain project lists only the root.
+// Only meaningful for yarn v2+; v1 does not support the command.
+func IsYarnWorkspaceProject(executablePath, srcPath string) (bool, error) {
```
agrasth, line 280: this says, `only meaningful for yarn v2+; v1 does not support the command.` but `getDependenciesMap` calls it unconditionally. If the project runs Yarn v1, yarn workspaces list will fail and emit a Warn, silently skipping workspace seeding.
maybe checking `isV2AndAbove` before calling this, same way `GetYarnDependenciesWithVersion` does it internally, wdyt?

build/utils/yarn.go
```diff
@@ -264,6 +272,30 @@ func buildYarnV2DependencyMap(packageInfo *PackageInfo, responseStr string) (dep
...
+	command := exec.Command(executablePath, "workspaces", "list", "--json")
+	command.Dir = srcPath
+	outBuffer := bytes.NewBuffer([]byte{})
```
agrasth: Can be simplified to `new(bytes.Buffer)` or just `var outBuffer bytes.Buffer` (and pass `&outBuffer`). Same for errBuffer.
Or it this intensional for some reason?

build/utils/yarn.go
```diff
@@ -410,3 +442,53 @@ type YarnDependencyPointer struct {
...
+// nonRegistryProtocols lists yarn locator protocols that resolve from local disk or
+// VCS rather than an npm registry / Artifactory.
+var nonRegistryProtocols = map[string]bool{
+	"workspace": true,
+	"link":      true,
+	"file":      true,
+	"portal":    true,
+	"patch":     true,
+	"git+https": true,
+	"git+ssh":   true,
```
agrasth, line 479: `git+ssh `is listed in `nonRegistryProtocols` but there's no test case for it in `TestIsNonRegistryLocator`. Worth adding `"pkg@git+ssh://github.com/org/repo"` to the non-registry cases.

agrasth (approved):

PR #380, opened 2026-05-12
RTECO-1265: fix(dotnet): preserve all per-TFM package versions in build-info dependency collection

Edenb2131:
## Problem

For .NET projects with multiple target frameworks (e.g. `<TargetFrameworks>net8.0;net10.0</TargetFrameworks>`) and TFM-conditional `PackageReference` versions, `jf dotnet restore` + `jf dotnet build` + `jf rt build-publish` produce a build-info that silently drops dependency versions.

**Root cause:** `assetsjson.go` stores library entries from `project.assets.json` into Go maps keyed by **package name only**. When the same package resolves to different versions across TFMs, entries overwrite each other. Go's randomized map iteration order determines which version survives — making the result nondeterministic across runs. All three map-producing functions were affected: `getAllDependencies`, `getChildrenMap`, and `getDirectDependencies`.

**Impact on Xray:** Dropped versions are never scanned, meaning known CVEs in those versions go unflagged. This directly undermines the customer's software supply chain security posture.

## Fix

Key all three dependency maps on `name:version` instead of `name` only, preserving every per-TFM `(name, version)` pair:

- `getAllDependencies` — map key changed from `getDependencyName(id)` → `strings.ToLower(getDependencyIdForBuildInfo(id))`
- `getChildrenMap` — parent key changed to `name:version`; child values now include version from `targetDependency.Dependencies` map values; sets used to deduplicate across TFMs
- `getDirectDependencies` — cross-references `Libraries` to resolve `name:version` for each direct dep name; sorted for deterministic output

## Tests

- Updated `TestNewAssetsExtractor`, `TestGetDirectDependenciesDeterministic`, `TestGetChildrenMapDeterministic` for the new key format
- Added `TestMultiTFMVersionPreservation` regression test with a new testdata fixture (2 packages × 2 TFMs = 4 entries — previously only 2 survived nondeterministically)
- Added `TestGetDependencyIdForBuildInfo` and `TestSetToSortedSlice`
- Updated `TestLoadMixed` in the `solution` package

## Reproduction

A full reproduction pipeline demonstrating the bug (3 parallel runs to surface nondeterminism) and a control pipeline (plain `dotnet`, no JFrog CLI, confirming `project.assets.json` is always correct) are available at:
https://github.com/Edenb2131/dotnet-multitarget-buildinfo-repro

Jira: RTECO-1265

### DIFF
--- a/build/utils/dotnet/dependencies/assetsjson.go
+++ b/build/utils/dotnet/dependencies/assetsjson.go
@@ -67,23 +67,24 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log) 
 }
 
 func (assets *assets) getChildrenMap() map[string][]string {
-	// Use a set to deduplicate children across multiple target frameworks
+	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs
 	dependenciesRelations := map[string]map[string]struct{}{}
 	for _, dependencies := range assets.Targets {
 		for dependencyId, targetDependencies := range dependencies {
-			dependencyName := getDependencyName(dependencyId)
-			if _, ok := dependenciesRelations[dependencyName]; !ok {
-				dependenciesRelations[dependencyName] = map[string]struct{}{}
+			dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
+			if _, ok := dependenciesRelations[dependencyKey]; !ok {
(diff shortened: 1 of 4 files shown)


Edenb2131:

Edenb2131:
recheck

Edenb2131:
Closing this as we have issue with C:A. Going to create a new one.

PR #381, opened 2026-05-13
RTECO-1265: fix(dotnet): preserve all per-TFM package versions in build-info

Edenb2131:
### **Problem**
For .NET projects with multiple target frameworks (e.g. <TargetFrameworks>net8.0;net10.0</TargetFrameworks>) and TFM-conditional PackageReference versions, jf dotnet restore + jf dotnet build + jf rt build-publish produce a build-info that silently drops dependency versions.

**Root cause:** assetsjson.go stores library entries from project.assets.json into Go maps keyed by package name only. When the same package resolves to different versions across TFMs, entries overwrite each other. Go's randomized map iteration order determines which version survives — making the result nondeterministic across runs. All three map-producing functions were affected: getAllDependencies, getChildrenMap, and getDirectDependencies.

**Impact on Xray:** Dropped versions are never scanned, meaning known CVEs in those versions go unflagged. This directly undermines the customer's software supply chain security posture.

### **Fix**
Key all three dependency maps on name:version instead of name only, preserving every per-TFM (name, version) pair:

getAllDependencies — map key changed from getDependencyName(id) → strings.ToLower(getDependencyIdForBuildInfo(id))
getChildrenMap — parent key changed to name:version; child values now include version from targetDependency.Dependencies map values; sets used to deduplicate across TFMs
getDirectDependencies — cross-references Libraries to resolve name:version for each direct dep name; sorted for deterministic output.
Note: two intentional behavioral changes: 
1. Type == "project" entries (local project refs, not NuGet packages) are excluded, consistent with AllDependencies()
2. Framework deps not present in Libraries are excluded; this is impossible after a successful dotnet restore.

### **Tests**
Updated TestNewAssetsExtractor, TestGetDirectDependenciesDeterministic, TestGetChildrenMapDeterministic for the new key format
Added TestMultiTFMVersionPreservation regression test with a new testdata fixture (2 packages × 2 TFMs = 4 entries — previously only 2 survived nondeterministically)
Added TestGetDependencyIdForBuildInfo and TestSetToSortedSlice
Updated TestLoadMixed in the solution package
### **Reproduction**
A full reproduction pipeline demonstrating the bug (3 parallel runs to surface nondeterminism) and a control pipeline (plain dotnet, no JFrog CLI, confirming project.assets.json is always correct) are available at:
https://github.com/Edenb2131/dotnet-multitarget-buildinfo-repro

Jira: RTECO-1265

### DIFF
--- a/build/utils/dotnet/dependencies/assetsjson.go
+++ b/build/utils/dotnet/dependencies/assetsjson.go
@@ -76,7 +76,7 @@ func (assets *assets) getChildrenMap() map[string][]string {
 				dependenciesRelations[dependencyKey] = map[string]struct{}{}
 			}
 			for transitiveName, transitiveVersion := range targetDependencies.Dependencies {
-				childKey := strings.ToLower(transitiveName) + ":" + transitiveVersion
+				childKey := strings.ToLower(transitiveName + ":" + transitiveVersion)
 				dependenciesRelations[dependencyKey][childKey] = struct{}{}
 			}
 		}
--- a/build/utils/dotnet/dependencies/assetsjson.go
+++ b/build/utils/dotnet/dependencies/assetsjson.go
@@ -67,23 +67,24 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log) 
 }
 
 func (assets *assets) getChildrenMap() map[string][]string {
-	// Use a set to deduplicate children across multiple target frameworks
+	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs
 	dependenciesRelations := map[string]map[string]struct{}{}
 	for _, dependencies := range assets.Targets {
 		for dependencyId, targetDependencies := range dependencies {
-			dependencyName := getDependencyName(dependencyId)
-			if _, ok := dependenciesRelations[dependencyName]; !ok {
-				dependenciesRelations[dependencyName] = map[string]struct{}{}
+			dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
+			if _, ok := dependenciesRelations[dependencyKey]; !ok {
+				dependenciesRelations[dependencyKey] = map[string]struct{}{}
 			}
-			for transitiveName := range targetDependencies.Dependencies {
-				dependenciesRelations[dependencyName][strings.ToLower(transitiveName)] = struct{}{}
+			for transitiveName, transitiveVersion := range targetDependencies.Dependencies {
+				childKey := strings.ToLower(transitiveName) + ":" + transitiveVersion
+				dependenciesRelations[dependencyKey][childKey] = struct{}{}
 			}
 		}
 	}
 	// Convert sets to sorted slices for deterministic output
 	result := make(map[string][]string, len(dependenciesRelations))
-	for dependencyName, transitiveSet := range dependenciesRelations {
-		result[dependencyName] = setToSortedSlice(transitiveSet)
+	for dependencyKey, transitiveSet := range dependenciesRelations {
+		result[dependencyKey] = setToSortedSlice(transitiveSet)
 	}
 	return result
 }
@@ -98,15 +99,25 @@ func setToSortedSlice(values map[string]struct{}) []string {
 }
 
 func (assets *assets) getDirectDependencies() []string {
-	// Use a set to deduplicate across multiple target frameworks
-	directDependencies := map[string]struct{}{}
+	// Collect direct dep names from all frameworks
+	directNames := map[string]bool{}
 	for _, framework := range assets.Project.Frameworks {
-		for dependencyName := range framework.Dependencies {
-			directDependencies[strings.ToLower(dependencyName)] = struct{}{}
+		for depName := range framework.Dependencies {
+			directNames[strings.ToLower(depName)] = true
 		}
 	}
-	// Return sorted slice for deterministic output
-	return setToSortedSlice(directDependencies)
+	// Cross-reference with Libraries to resolve name:version for each direct dep.
+	// A single package name may resolve to multiple versions across TFMs — each is kept.
+	seen := map[string]struct{}{}
+	for libId, library := range assets.Libraries {
+		if library.Type == "project" {
+			continue
+		}
+		if directNames[getDependencyName(libId)] {
+			seen[strings.ToLower(getDependencyIdForBuildInfo(libId))] = struct{}{}
+		}
+	}
+	return setToSortedSlice(seen)
 }
 
 func (assets *assets) getAllDependencies(log utils.Log) (map[string]*buildinfo.Dependency, error) {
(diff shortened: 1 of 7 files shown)


Edenb2131:

build/utils/dotnet/dependencies/assetsjson.go
```diff
@@ -67,23 +67,24 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log)
...
 func (assets *assets) getChildrenMap() map[string][]string {
-	// Use a set to deduplicate children across multiple target frameworks
+	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs
 	dependenciesRelations := map[string]map[string]struct{}{}
 	for _, dependencies := range assets.Targets {
 		for dependencyId, targetDependencies := range dependencies {
-			dependencyName := getDependencyName(dependencyId)
-			if _, ok := dependenciesRelations[dependencyName]; !ok {
-				dependenciesRelations[dependencyName] = map[string]struct{}{}
+			dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
```
naveenku-jfrog, line 74: Switching the parent key from name to name:version here breaks solution.populateRequestedBy, which looks up childrenMap[getDependencyName(parent.Id)] (name only). Result: every transitive dep loses its RequestedBy and gets filtered out of the build-info.

build/utils/dotnet/dependencies/assetsjson.go
```diff
@@ -67,23 +67,24 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log)
...
+			if _, ok := dependenciesRelations[dependencyKey]; !ok {
+				dependenciesRelations[dependencyKey] = map[string]struct{}{}
 			}
-			for transitiveName := range targetDependencies.Dependencies {
-				dependenciesRelations[dependencyName][strings.ToLower(transitiveName)] = struct{}{}
+			for transitiveName, transitiveVersion := range targetDependencies.Dependencies {
+				childKey := strings.ToLower(transitiveName) + ":" + transitiveVersion
```
naveenku-jfrog: Children stored as name:version will only match AllDependencies keys, not populateRequestedBy's name-only lookups — please update solution.go's traversal in the same PR.

build/utils/dotnet/dependencies/assetsjson.go
```diff
@@ -98,15 +99,25 @@ func setToSortedSlice(values map[string]struct{}) []string {
...
-	return setToSortedSlice(directDependencies)
+	// Cross-reference with Libraries to resolve name:version for each direct dep.
+	// A single package name may resolve to multiple versions across TFMs — each is kept.
+	seen := map[string]struct{}{}
+	for libId, library := range assets.Libraries {
+		if library.Type == "project" {
+			continue
+		}
+		if directNames[getDependencyName(libId)] {
+			seen[strings.ToLower(getDependencyIdForBuildInfo(libId))] = struct{}{}
```
naveenku-jfrog, line 117: Deriving direct deps from Libraries silently drops any framework dependency that isn't in Libraries and excludes Type == "project" refs that the old code returned — please confirm intentional and call out in the PR description.

build/utils/dotnet/dependencies/assetsjson_test.go
```diff
@@ -191,14 +198,41 @@ func TestGetChildrenMapDeterministic(t *testing.T) {
 	assert.NoError(t, json.Unmarshal(content, &assetsObj))
```
naveenku-jfrog, line 200: This test only validates the three extractor maps in isolation. Please add an end-to-end assertion on solution.BuildInfo() for testdata/multitfm — that's where the real regression (transitive deps dropped) would surface.

build/utils/dotnet/solution/solution_test.go
```diff
@@ -401,12 +401,12 @@ func TestLoadMixed(t *testing.T) {
 			assert.Equal(t, filepath.Join(wd, "tmp", "multi", "multi"), project.RootPath())
 			direct, err := project.Extractor().DirectDependencies()
 			require.NoError(t, err)
-			assert.ElementsMatch(t, []string{"newtonsoft.json"}, direct)
+			assert.ElementsMatch(t, []string{"newtonsoft.json:9.0.1"}, direct)
```
naveenku-jfrog, line 425: TestPopulateRequestedByDeterministic above still uses name-only keys ("depa", "depb", …) and isn't updated to the new name:version shape — it no longer reflects what assetsExtractor produces.

naveenku-jfrog (changes requested):

build/utils/dotnet/dependencies/assetsjson.go
Edenb2131: That means the function getDependencyName() is no longer relevant in solution.go (never called). Can delete it?

build/utils/dotnet/dependencies/assetsjson.go
naveenku-jfrog: The version isn't lowercased here, but the parent key (line 76) and the AllDependencies key (line 151) both lowercase the full name:version. For packages with mixed-case prerelease labels (e.g. 1.0.0-Beta, -RC1, -Preview.1), transitiveVersion can be written in non-normalized form inside targets[...].dependencies (it comes from the consuming package's .nuspec), so childrenMap would store "pkg:1.0.0-Beta" while AllDependencies stores "pkg:1.0.0-beta" — populateRequestedBy's dependenciesMap[childName] lookup then misses and the child gets silently dropped from build-info. Same failure mode as the bug this PR is fixing.

naveenku-jfrog (changes requested):

naveenku-jfrog (approved):

Edenb2131:

JFrogEden:

build/utils/dotnet/solution/solution.go
```diff
@@ -124,7 +124,7 @@ func getModuleId(customModuleID, projectName string) string {
 // dependenciesMap  - The input dependencies map
 // childrenMap      - Map from dependency ID to children IDs
 func populateRequestedBy(parentDependency buildinfo.Dependency, dependenciesMap map[string]*buildinfo.Dependency, childrenMap map[string][]string) {
-	childrenList := childrenMap[getDependencyName(parentDependency.Id)]
+	childrenList := childrenMap[strings.ToLower(parentDependency.Id)]
```
bhanurp: This change breaks RequestedBy for .NET Framework / packages.config projects(edge case for legacy packages.config-only projects).

packagesExtractor (unchanged in this PR) keys childrenMap by name only — see [packagesconfig.go:75,99] This is the scenario when it breaks
  - A packages.config is present, AND
  - No obj/project.assets.json exists for that project, AND
  - The directory-matching quirk above doesn't pull in a sibling project's assets file.

Suggested fix: 
```
key := strings.ToLower(parentDependency.Id)
  childrenList, ok := childrenMap[key]
  if !ok {
      // Legacy fallback: packagesExtractor still keys by name only.
      // Remove this once packagesExtractor is migrated to name:version.
      if idx := strings.Index(key, ":"); idx != -1 {
          childrenList = childrenMap[key[:idx]]
      }
  }
```

bhanurp (approved):

PR #385, opened 2026-05-21
RTECO-1265 - Fixed Nuget bracket based dependency resolution

fluxxBot:
-----

## Summary

  Fix `getChildrenMap` in NuGet `project.assets.json` extractor so transitive dependencies survive `populateRequestedBy` when parent packages declare children
  with bracket-range version notation.

  ## What was wrong

  PR #381 (`RTECO-1265: preserve all per-TFM package versions`) rekeyed the NuGet dependency maps from `name` to `name:version` to support multi-TFM resolution.
  The implementation built `childrenMap` child keys from the **declared constraint string** found in `targets[tfm][parentId].dependencies`:

  ```go
  childKey := strings.ToLower(transitiveName + ":" + transitiveVersion)

  That value is whatever the parent's nuspec declared — sometimes a plain version ("1.0.1"), sometimes a range ("[1.12.9, )"). The dependencies map
  (getAllDependencies) is keyed by the resolved library version. When a parent used range notation, the two keys diverged:

  - childrenMap["bootstrap:4.0.0"] = ["jquery:[3.0.0, )", "popper.js:[1.12.9, )"]
  - dependenciesMap = [..., "jquery:3.0.0", "popper.js:1.12.9", ...]

  populateRequestedBy looked up dependenciesMap[childName] with childName = "popper.js:[1.12.9, )" → miss → child stayed with empty RequestedBy →
  solution.BuildInfo filtered it out via the len(dep.RequestedBy) > 0 guard.

  Net effect: any transitive whose parent declared a version range was dropped from the published build-info. Caught by TestNugetResolve / TestDotnetResolve in
  jfrog-cli, where bootstrap 4.0.0 declares jQuery + popper.js with brackets — both deps disappeared from every reference/multireference scenario.

  The existing unit tests passed because every fixture (assetsproject, multitfm, the synthetic TestGetChildrenMapDeterministic data) used plain version strings
  only — bracket-range case was uncovered.

  What this fixes

  getChildrenMap now resolves the child key against the same target's library entries:

  1. Per TFM, build resolvedInTfm[name] = "name:resolvedVersion" from each library id in that target.
  2. For each transitive dep, look up resolvedInTfm[transitiveName] and use that as childKey.
  3. If no library entry exists in the TFM (synthetic / malformed assets), fall back to declared name:version to preserve existing behavior.

  Result: childrenMap keys and dependenciesMap keys agree on the resolved version. populateRequestedBy finds every child. No transitive deps dropped.

  Multi-TFM correctness from PR #381 is preserved — resolvedInTfm is rebuilt per target, so different TFMs resolving the same name to different versions still
  produce distinct child keys.

### DIFF
--- a/build/utils/dotnet/dependencies/assetsjson.go
+++ b/build/utils/dotnet/dependencies/assetsjson.go
@@ -67,16 +67,31 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log) 
 }
 
 func (assets *assets) getChildrenMap() map[string][]string {
-	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs
+	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs.
+	// Transitive version strings in targets/dependencies are the declared constraint (e.g. "[1.12.9, )"),
+	// not the resolved version. Build a per-TFM name->resolved-name:version map so child keys match
+	// the resolved versions used as keys in dependenciesMap (getAllDependencies).
 	dependenciesRelations := map[string]map[string]struct{}{}
 	for _, dependencies := range assets.Targets {
+		resolvedInTfm := map[string]string{}
+		for depId := range dependencies {
+			if idx := strings.Index(depId, "/"); idx != -1 {
+				resolvedInTfm[strings.ToLower(depId[:idx])] = strings.ToLower(getDependencyIdForBuildInfo(depId))
+			}
+		}
 		for dependencyId, targetDependencies := range dependencies {
 			dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
 			if _, ok := dependenciesRelations[dependencyKey]; !ok {
 				dependenciesRelations[dependencyKey] = map[string]struct{}{}
 			}
 			for transitiveName, transitiveVersion := range targetDependencies.Dependencies {
-				childKey := strings.ToLower(transitiveName + ":" + transitiveVersion)
+				// Prefer per-TFM resolved version (from library entry in the same target).
+				// Fall back to declared constraint when no library entry exists in this TFM —
+				// rare in real assets.json but kept for safety/back-compat.
+				childKey, ok := resolvedInTfm[strings.ToLower(transitiveName)]
+				if !ok {
+					childKey = strings.ToLower(transitiveName + ":" + transitiveVersion)
+				}
 				dependenciesRelations[dependencyKey][childKey] = struct{}{}
 			}
 		}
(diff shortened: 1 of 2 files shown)


build/utils/dotnet/dependencies/assetsjson.go
bhanurp, line 1: shouldnt we point this to dev?

build/utils/dotnet/dependencies/assetsjson.go
```diff
@@ -67,16 +67,31 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log)
...
+	// Key by name:version to preserve per-TFM entries; use a set to deduplicate children across TFMs.
+	// Transitive version strings in targets/dependencies are the declared constraint (e.g. "[1.12.9, )"),
+	// not the resolved version. Build a per-TFM name->resolved-name:version map so child keys match
+	// the resolved versions used as keys in dependenciesMap (getAllDependencies).
 	dependenciesRelations := map[string]map[string]struct{}{}
 	for _, dependencies := range assets.Targets {
+		resolvedInTfm := map[string]string{}
+		for depId := range dependencies {
+			if idx := strings.Index(depId, "/"); idx != -1 {
+				resolvedInTfm[strings.ToLower(depId[:idx])] = strings.ToLower(getDependencyIdForBuildInfo(depId))
```
bhanurp, line 80: lets add unit tests for bracket range so that regressions can be avoided

build/utils/dotnet/dependencies/assetsjson.go
```diff
@@ -67,16 +67,31 @@ func (extractor *assetsExtractor) new(dependenciesSource string, log utils.Log)
...
+		}
 		for dependencyId, targetDependencies := range dependencies {
 			dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
 			if _, ok := dependenciesRelations[dependencyKey]; !ok {
 				dependenciesRelations[dependencyKey] = map[string]struct{}{}
 			}
 			for transitiveName, transitiveVersion := range targetDependencies.Dependencies {
-				childKey := strings.ToLower(transitiveName + ":" + transitiveVersion)
+				// Prefer per-TFM resolved version (from library entry in the same target).
+				// Fall back to declared constraint when no library entry exists in this TFM —
```
bhanurp, line 90: lets a debug log to tell that there is a fallback, when it fallsback and misses any deps it will be hard to debug otherwise

build/utils/dotnet/dependencies/assetsjson.go
fluxxBot: no we do not release build-info-go anymore, changes are in supposed to be directly in main

bhanurp (approved):

PR #392, opened 2026-06-15
RTECO-1187 - Updated logic to skip cache check for empty depedency project

naveenku-jfrog:
-----
What: Fix jf npm ci (and any caller of build.CollectDependenciesAndPublishBuildInfo) failing with _cacache folder is not found in '<npm-cache>/_cacache' on projects whose package.json declares no dependencies.

Why: npm only creates <npm-cache>/_cacache after it fetches its first tarball, so a zero-dependency project on a fresh CI workspace (e.g. a Jenkins agent that wipes between builds) never has it — but CalculateNpmDependenciesList was unconditionally requiring it for checksum calculation, hard-failing builds that have nothing to checksum in the first place.

How: Guard the GetNpmConfigCache call in build/utils/npm.go on len(dependenciesMap) > 0 so the _cacache lookup is skipped when there are no dependencies; added TestCalculateNpmDependenciesListWithoutDependencies as a regression test (verified failing on main, passing with the fix).

### DIFF
--- a/build/utils/npm.go
+++ b/build/utils/npm.go
@@ -33,8 +33,11 @@ func CalculateNpmDependenciesList(executablePath, srcPath, moduleId string, npmP
 		return nil, err
 	}
 	var cacache *cacache
-	if calculateChecksums {
-		// Get local npm cache.
+	// Skip the _cacache lookup when there are no dependencies to checksum.
+	// A project with no dependencies never causes npm to create the _cacache
+	// directory, so requiring it would fail `jf npm ci` on a fresh CI workspace
+	// (e.g. a Jenkins agent) for projects whose package.json has no dependencies.
+	if calculateChecksums && len(dependenciesMap) > 0 {
 		cacheLocation, err := GetNpmConfigCache(srcPath, executablePath, npmParams.Args, log)
 		if err != nil {
 			return nil, err
--- a/build/utils/npm_test.go
+++ b/build/utils/npm_test.go
@@ -461,18 +461,27 @@ func TestCalculateNpmDependenciesListWithoutDependencies(t *testing.T) {
 	cachePath := filepath.Join(tempDir, "tmpcache")
 	npmArgs := []string{"--cache=" + cachePath}
 
-	// Generate package-lock.json (required by 'npm ci') without populating _cacache.
+	// Generate package-lock.json (required by 'npm ci' in real-world usage) without
+	// installing any tarballs. With modern npm this leaves _cacache absent; older npm
+	// (e.g. npm v6 bundled with Node 14) creates _cacache eagerly even on a no-op
+	// install. We force the "missing _cacache" state below to make the test
+	// deterministic across npm versions and to faithfully reproduce a fresh CI
+	// workspace (e.g. a Jenkins agent that wipes node_modules + the npm cache).
 	installArgs := append([]string{"--package-lock-only"}, npmArgs...)
 	_, _, err := RunNpmCmd("npm", tempDir, AppendNpmCommand(installArgs, "install"), logger)
 	require.NoError(t, err)
 
-	// Precondition: the npm cache must NOT contain _cacache. If this ever changes
-	// (e.g. a future npm version starts creating it eagerly), the test would silently
-	// stop exercising the regression path; fail loudly instead.
 	cacachePath := filepath.Join(cachePath, "_cacache")
+	if exists, err := utils.IsDirExists(cacachePath, false); err == nil && exists {
+		require.NoError(t, os.RemoveAll(cacachePath))
+	}
+
+	// Post-condition: _cacache must be absent. This is the precondition for the
+	// regression scenario; if it ever fails, the test would silently stop
+	// exercising the new guard, so fail loudly instead.
 	cacacheExists, err := utils.IsDirExists(cacachePath, false)
 	require.NoError(t, err)
-	require.Falsef(t, cacacheExists, "test precondition violated: '_cacache' must not exist at %q for this regression test to be meaningful", cacachePath)
+	require.Falsef(t, cacacheExists, "could not remove _cacache at %q; regression test would not exercise the fixed code path", cacachePath)
 
 	// The actual regression check.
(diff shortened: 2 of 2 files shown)


reshmifrog (approved):

build/utils/npm_test.go
```diff
@@ -427,6 +427,69 @@ func TestGetConfigCacheNpmIntegration(t *testing.T) {
...
+//     precondition for 'npm ci') but does NOT create '_cacache' because no tarballs
+//     are fetched.
+//
+// Before the fix, CalculateNpmDependenciesList returned the error:
+//
+//	"_cacache folder is not found in '<cache>/_cacache'. Hint: Delete node_modules
+//	 directory and run npm install or npm ci."
+//
+// After the fix it returns no error and an empty dependency list.
+func TestCalculateNpmDependenciesListWithoutDependencies(t *testing.T) {
```
agrasth, line 448: **Minor** — This test calls `RunNpmCmd("npm", ...)` which requires a live `npm` binary, making it an integration test. The neighbouring `TestGetConfigCacheNpmIntegration` follows the apparent project convention of using an `Integration` suffix for npm-CLI-dependent tests; this one does not. If CI runs unit tests in an environment without `npm`, the failure will be an obscure exec error rather than a clear categorisation signal.

Suggested rename: `TestCalculateNpmDependenciesListWithoutDependenciesIntegration`

build/utils/npm.go
```diff
@@ -33,8 +33,11 @@ func CalculateNpmDependenciesList(executablePath, srcPath, moduleId string, npmP
 		return nil, err
 	}
 	var cacache *cacache
-	if calculateChecksums {
-		// Get local npm cache.
+	// Skip the _cacache lookup when there are no dependencies to checksum.
```
agrasth, line 36: **Nit** — Lines 2-4 of this comment block explain Jenkins CI workspace semantics that belong in the commit message or PR description, not in the source file. They will drift from reality over time. The first line alone captures the invariant:

```go
// Skip the _cacache lookup when there are no dependencies to checksum.
```

agrasth (commented):
## Ticket Alignment

- **Implements the ticket requirement**: Yes — the guard `len(dependenciesMap) > 0` in `CalculateNpmDependenciesList` correctly skips the `_cacache` lookup when there are no dependencies, preventing the hard failure described in RTECO-1187 on projects with empty `package.json` dependencies on fresh CI workspaces.
- **Missing**: Nothing material — the fix matches the described root cause and expected behaviour exactly.
- **Out of scope**: Nothing in this PR exceeds the ticket's scope.

---

## Summary

This is a clean, minimal, correct fix. The one-line production change is safe: when `dependenciesMap` is empty the loop over it never executes, so the nil `cacache` is never dereferenced. The regression test is thorough and deterministic across npm versions.

One minor finding (test naming) and two nits are noted below; none block merging.

---

## Findings

### Testing

#### Minor — `build/utils/npm_test.go:448` — Test name should carry "Integration" suffix

This test calls `RunNpmCmd("npm", ...)` and therefore requires a live `npm` binary. The neighbouring `TestGetConfigCacheNpmIntegration` carries the `Integration` suffix as a convention for npm-CLI-dependent tests. Without this suffix, if CI runs unit tests in an environment without `npm`, this test will fail with a confusing error (e.g. `exec: "npm": executable file not found in $PATH`) rather than being skipped or clearly labelled as infrastructure-dependent.

Suggested name: `TestCalculateNpmDependenciesListWithoutDependenciesIntegration`

### Maintainability

#### Nit — `build/utils/npm.go:36` — 4-line comment over-explains; first line is sufficient

The added comment block explains Jenkins CI workspace behaviour (lines 2-4) that belongs in the commit message or PR description rather than the source file. Lines 2-4 may drift from reality as the codebase evolves. The first line captures the invariant; the rest is implementation context.

Suggested: keep only the first line:
```go
// Skip the _cacache lookup when there are no dependencies to checksum.
```

#### Nit — `build/utils/npm_test.go:475` — Conditional `os.RemoveAll` guard is unnecessary

`os.RemoveAll` is a no-op when the path does not exist, so the `if exists, err := ...; err == nil && exists { ... }` wrapper is unnecessary. An unconditional call is simpler and also surfaces any removal error directly:
```go
require.NoError(t, os.RemoveAll(cacachePath))
```

---

## Positive Highlights

- **Minimal blast radius**: a single guard condition added; all existing behaviour is preserved when `dependenciesMap` is non-empty.
- **Nil-safe by construction**: `cacache` remains nil only when the loop body will never execute (empty map), so there is no nil dereference risk.
- **Deterministic regression test**: the test forces `_cacache` absent regardless of npm version, then asserts its absence loudly before invoking the function under test — a well-structured regression guard.
- **Strong post-condition message**: `require.Falsef` failure message clearly tells a future maintainer why the test exists and what path it is guarding.

build/utils/npm_test.go
```diff
@@ -427,6 +427,69 @@ func TestGetConfigCacheNpmIntegration(t *testing.T) {
...
+	// (e.g. npm v6 bundled with Node 14) creates _cacache eagerly even on a no-op
+	// install. We force the "missing _cacache" state below to make the test
+	// deterministic across npm versions and to faithfully reproduce a fresh CI
+	// workspace (e.g. a Jenkins agent that wipes node_modules + the npm cache).
+	installArgs := append([]string{"--package-lock-only"}, npmArgs...)
+	_, _, err := RunNpmCmd("npm", tempDir, AppendNpmCommand(installArgs, "install"), logger)
+	require.NoError(t, err)
+
+	cacachePath := filepath.Join(cachePath, "_cacache")
+	if exists, err := utils.IsDirExists(cacachePath, false); err == nil && exists {
```
agrasth, line 475: **Nit** — `os.RemoveAll` is a no-op when the path does not exist, so the guard is unnecessary. An unconditional call is simpler and also surfaces any removal error directly:

```go
require.NoError(t, os.RemoveAll(cacachePath))
```

agrasth (approved):

PR #393, opened 2026-06-19
RTECO-945: Add Alpine APK build-info collection support

naveenku-jfrog:
## Summary

Adds native Alpine Linux APK build-info collection to `build-info-go`, enabling `bi apk` to record installed packages as structured Build Info dependencies with checksums, scopes, and `requestedBy` chains — consistent with how other package managers (npm, pip, Maven) are supported in this library.

---

### New: `build/utils/alpine.go`
Core utilities for Alpine package handling:
- **`AlpinePackage` struct** — holds `Name`, `Version`, `Arch`, `Size`, `Origin`, `URL`, and `Checksum` (raw `C:` field from the APK database)
- **`SHA1Hex()`** — decodes the Alpine DB checksum field (`Q1<base64>`) into a standard hex SHA1 string, available on every Alpine system without reading any files
- **`ListInstalledPackages()` / `parseInstalledDB()`** — parses `/lib/apk/db/installed` directly (zero subprocesses), returning deduplicated packages with full metadata including the `C:` checksum field
- **`DiffAlpinePackages()`** — snapshot-diff model consistent with npm/pip modules in this library
- **`ChecksumsFromCache()`** — computes SHA-256, MD5, and SHA-1 from cached `.apk` archives in `/var/cache/apk` (fallback; only populated when APK cache is explicitly enabled)
- **`BuildDepGraph()` / `parseDependsOutput()`** — builds a forward dependency graph via `apk info --depends` for `requestedBy` chain construction

### New: `build/alpine.go`
The `AlpineModule` that integrates into the build pipeline:
- **`AlpineModule`** — attached to a `Build`, orchestrates pre/post snapshot diffing to identify newly installed packages
- **`SetPreSnapshot()` / `SetCacheDir()` / `SetRequestedPackages()`** — configure the module before `apk add` runs
- **`CollectDependencies()`** — diffs snapshots, resolves checksums (DB-primary, file-cache fallback), classifies scope (`prod` vs `transitive`), and builds `requestedBy` chains
- **`SaveBuildInfo()`** — persists the collected dependencies as an `entities.Alpine` module
- **`buildRequestedBy()`** — populates `RequestedBy` chains with a depth guard (`RequestedByMaxLength = 15`)

### Modified: `build/build.go`, `entities/buildinfo.go`, `cli/cli.go`
- Registered `AlpineModule` in the build factory via `AddAlpineModule()`
- Added `Alpine` package type constant (`entities.Alpine = "alpine"`)
- CLI wiring for the `jf bi apk` command

### New: `build/alpine_test.go`, `build/utils/alpine_test.go`
34 unit tests covering:
- `AlpinePackage.ID()` format (`name:version`)
- `SHA1Hex()` — valid, empty, and malformed checksum fields
- `parseInstalledDB()` — including `C:` field parsing and deduplication
- `DiffAlpinePackages()` — all edge cases (empty before/after, duplicates, multiple new packages)
- `ChecksumsFromCache()` — cache hit and miss
- `BuildDepGraph()` / `parseDependsOutput()` — version constraint stripping, virtual provider prefixes
- `buildRequestedBy()` — direct packages, transitive deps, multiple parents, depth cap
- `AlpineModule` setters, `SaveBuildInfo` module type, end-to-end scope classification

---

## Design Notes

**Checksum strategy**: The `C:` field in `/lib/apk/db/installed` is populated for every package on every Alpine system (Docker, VM, bare metal) immediately after `apk add`. Using it as the primary SHA-1 source means checksums are always available without relying on `/var/cache/apk`, which is opt-in and absent in most Docker containers. SHA-256 and MD5 are additionally computed from the cached `.apk` archive when available.

**No subprocess dependency**: Package metadata is parsed directly from `/lib/apk/db/installed` in Go — no runtime dependency on the `apk` binary for metadata collection.

**Snapshot-diff model**: Consistent with how npm and pip work in this library — snapshot before `apk add`, snapshot after, diff gives the newly installed set.

**Dependency ID format**: `name:version` (e.g. `curl:8.14.1-r2`), consistent with the convention used by npm, pip, and other modules in this library.


naveenku-jfrog (commented):
Testing PR review through agent-workbench

PR #394, opened 2026-06-29
RTECO-1536 - Add RubyGems FlexPack for native build-info collection

agrasth:
## Summary

Adds RubyGems FlexPack support for native build-info collection from `Gemfile.lock`, and passes through the `Repository` field to `entities.Dependency` for enrichment by the CLI layer.

### What is included in this PR

- **`RubygemsFlexPack`**: Parses `Gemfile.lock` (GEM/GIT/PATH sections), resolves dependency graph, respects `InstalledPackages` filtering from `bundle list`
- **`GemConfig`**: Configuration struct with working directory and optional installed packages filter
- **`CollectBuildInfo`**: Produces `entities.BuildInfo` with module type `gem`, dependency IDs as `name:version`, and complete requestedBy chains
- **Repository field passthrough**: `dep.Repository` is now set on `entities.Dependency` during `CollectBuildInfo` so the CLI layer can enrich it from AQL results
- **Direct URL detection**: GIT/PATH gems marked with `DirectURL` to skip AQL enrichment
- **Module type**: `Gem ModuleType = "gem"` added to `entities/buildinfo.go`
- **7 unit tests** covering parsing, filtering, graph construction

## Test plan

- [x] `go test ./flexpack/ -run Rubygems -v` - all tests pass
- [x] `go build ./...` - compiles successfully


entities/buildinfo.go
```diff
@@ -353,11 +354,16 @@ func mergeDependenciesLists(dependenciesToAdd, intoDependencies *[]Dependency) {
 }
 
 func mergeDependencies(dep1, dep2 Dependency) Dependency {
+	repo := dep1.Repository
+	if repo == "" {
+		repo = dep2.Repository
```
reshmifrog, line 363: why are we doing this?

cli/cli.go
```diff
@@ -514,6 +514,26 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
 			},
 		},
+		{
+			Name:      "gem",
+			Usage:     "Generate build-info for a RubyGems project",
+			UsageText: "bi gem",
+			Flags:     flags,
+			Action: func(context *clitool.Context) (err error) {
+				config := flexpack.GemConfig{
+					WorkingDirectory: ".",
```
reshmifrog, line 573: Is it a validate case in which the config file is always present in the root directory.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+		}
+
+		// BUNDLED WITH holds a single indented version line.
+		if inBundledWith {
+			lock.bundlerVersion = trimmed
+			continue
+		}
+
+		// Inside a GEM/GIT/PATH block.
+		if currentSection != "" {
```
reshmifrog: can we use dependency tree graph command if available, that way it would be a lot easier.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	return name, version
+}
+
+// gemDependencyName extracts the gem name from a dependency line, stripping any
+// version constraint in parentheses and the trailing "!" pin marker.
+//
+//	"rspec-core (~> 3.12.0)" → "rspec-core"
+//	"rails (>= 6.0, < 7)"    → "rails"
+//	"my_gem!"                → "my_gem"
+func gemDependencyName(line string) string {
```
reshmifrog: why are we doing this?

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	rootChildren := rf.collectRootChildren(depInfoMap)
+
+	// Assign scopes from Gemfile groups when available.
+	if rf.config.GemGroups != nil {
+		rf.assignScopes(depInfoMap, fwdGraph, rootChildren)
+	}
+
+	// Reuse the shared chain builder (defined in uv_flexpack.go) — it operates purely
+	// on depInfoMap + fwdGraph keys, so exact gem names work the same as UV's normalised names.
+	buildUvRequestedBy(moduleID, []string{}, rootChildren, depInfoMap, fwdGraph, rf.requestedByChains, entities.RequestedByMaxLength)
```
reshmifrog: what is buildUvRequestedBy doing here?

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+
+	for _, dep := range depInfoMap {
+		rf.dependencies = append(rf.dependencies, *dep)
+	}
+}
```
reshmifrog, line 243: Please check whether we have pushed some other changes as well as part of this rubyGems implementation. Please revert incase it is there.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+		}
+		rf.propagateScopes(name, scopes, fwdGraph, depInfoMap, resolved)
+	}
+
+	// Apply resolved scopes to DependencyInfo.
+	for name, dep := range depInfoMap {
+		if scopes, ok := resolved[name]; ok {
+			dep.Scopes = scopes
+		} else {
+			dep.Scopes = []string{"production"}
```
reshmifrog: let's not explicitly give scopes names , let's use what is actual there in native command.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	if rf.projectVersion != "" {
+		return fmt.Sprintf("%s:%s", rf.projectName, rf.projectVersion)
+	}
+	return rf.projectName
+}
+
+// ===== FlexPackManager Interface =====
+
+// GetDependency returns a formatted string with dependency information.
+func (rf *RubygemsFlexPack) GetDependency() string {
```
reshmifrog, line 364: we can use a dependency graph command for this right, wdys?

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	for _, dep := range rf.dependencies {
+		if dep.DirectURL != "" {
+			result[dep.ID] = dep.DirectURL
+		}
+	}
+	return result
+}
+
+// GetDependencyGraph returns the complete dependency graph (ID → child IDs).
+func (rf *RubygemsFlexPack) GetDependencyGraph() (map[string][]string, error) {
```
reshmifrog, line 507: use a dependency graph command available in the native package manager and try to use .

reshmifrog (changes requested):

entities/buildinfo.go
agrasth: Fixed — added a comment explaining it: two `Dependency` records for the same ID can come from calls with different visibility into Artifactory (e.g. a `gem build` merge, which has none of its own, with a prior install that does) — keep whichever side actually resolved a repository.

cli/cli.go
agrasth: Checked — `WorkingDirectory: "."` matches every other package manager's debug action in this file (I counted 7: npm, pip, poetry, uv, nix, etc. all hardcode "."). Not something specific to the Ruby integration.

flexpack/rubygems_flexpack.go
agrasth: The doc comment directly above `gemDependencyName` already has three examples answering this (`"rspec-core (~> 3.12.0)" → "rspec-core"`, etc.) — let me know if it's still unclear and I'll expand it.

flexpack/rubygems_flexpack.go
agrasth: Fixed — renamed to `buildRequestedByChains` since it's shared by UV and RubyGems, not UV-specific (it operates purely on `depInfoMap`/`fwdGraph` keys).

flexpack/rubygems_flexpack.go
agrasth: Checked — the `"production"` default-scope label matches the project-wide convention already documented in `flexpack.go:129` and used the same way by the Poetry FlexPack. Not something specific to RubyGems, so kept as-is for consistency with the other package managers.

flexpack/rubygems_flexpack.go
agrasth: Same answer as the other "use a dependency graph command" comment on this file — this mirrors UV's lock-file-parsing design intentionally. Happy to revisit if you see a specific problem with it.

flexpack/rubygems_flexpack.go
agrasth: Same reasoning as the other two dependency-graph-command comments on this file — parsing `Gemfile.lock` directly mirrors UV's `uv.lock` handling, the closest sibling in this repo.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+			case trimmed == "specs:":
+				inSpecs = true
+				continue
+			case !inSpecs:
+				// revision:, ref:, branch:, glob:, etc. — metadata we don't need.
+				continue
+			}
+
+			// Within "specs:": indent 4 = spec, indent 6+ = that spec's dependency.
+			if indent <= 4 {
```
naveenku-jfrog: why hardcoded check on indent?

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+				}
+			}
+		}
+	}
+
+	return lock
+}
+
+// countLeadingSpaces returns the number of leading space characters.
+func countLeadingSpaces(s string) int {
```
naveenku-jfrog: can we use golang inbuilt function?

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+
+// parseSpecLine parses "name (version)" → name, version. The version may carry a
+// platform suffix, e.g. "nokogiri (1.13.9-x86_64-linux)"; the full string is kept.
+func parseSpecLine(line string) (name, version string) {
+	open := strings.Index(line, " (")
+	if open == -1 {
+		return strings.TrimSpace(line), ""
+	}
+	name = strings.TrimSpace(line[:open])
+	rest := line[open+2:]
```
naveenku-jfrog: it might cause out of index, as there is no check before the accessing open+2 index.

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	}
+	set := make(map[string]bool, len(a))
+	for _, s := range a {
+		set[s] = true
+	}
+	for _, s := range b {
+		if !set[s] {
+			return false
+		}
+	}
```
naveenku-jfrog, line 331: You have prepared a map for every entry in a, and then iterating over b by checking element, this will fail in below condition:
a=["hello","hello","hello","hey"], b=["hey","hey",hey",'hello"]

flexpack/rubygems_flexpack.go
```diff
@@ -0,0 +1,641 @@
...
+	return depList
+}
+
+// CalculateChecksum returns checksum maps for all dependencies.
+func (rf *RubygemsFlexPack) CalculateChecksum() []map[string]interface{} {
+	rf.ensureParsed()
+	var checksums []map[string]interface{}
+	for _, dep := range rf.dependencies {
+		checksums = append(checksums, map[string]interface{}{
```
naveenku-jfrog, line 390: why interface{}, we have checksum struct already.

naveenku-jfrog (changes requested):

flexpack/rubygems_flexpack.go
reshmifrog: This has been already discussed we should not explicitly name the scope of the dependency we should use the scope name what the native package manager uses.

flexpack/rubygems_flexpack.go
reshmifrog: Files parsing is something that can break the commands later on because of format change so please let's not do file parsing.And using a command that gives us the dependency graph is way better , wdyt?

reshmifrog (changes requested):

flexpack/rubygems_flexpack.go
agrasth: I checked what Bundler actually reports rather than what the other FlexPacks do: ungrouped gems come back as `default`, and `production` is a name Bundler never emits. So the label we were writing does not exist in the native tool at all.

```
addressable  groups=["default"]
rspec        groups=["development"]
```

Switching to the group names Bundler reports.

flexpack/rubygems_flexpack.go
agrasth: Good catch, this was a real bug.

`scopesEqual` built a set from the first slice and then only checked membership, so your example returned `true`: same length, same distinct values, different counts. A genuine scope update was being suppressed.

It now compares occurrence counts, and stays order insensitive because the merge path combines scopes through a map. Added an internal test using exactly your input, which fails against the previous implementation.

Fixed in 1e96021.

flexpack/rubygems_flexpack.go
agrasth: Checked this one and I don't think it can go out of range.

`open` is the index of `" ("`, so `open+2` is at most `len(line)`, and slicing `s[len(s):]` is legal in Go and yields an empty string rather than panicking. The case where the line ends in `" ("` is therefore handled.

That said, this function is going away entirely: the lock file parsing is being replaced with the dependency graph Bundler itself reports, so `parseSpecLine` is removed along with it.

flexpack/rubygems_flexpack.go
agrasth: Agreed in principle, but this signature is not ours to choose here.

`CalculateChecksum() []map[string]interface{}` is declared by the `FlexPackManager` interface in `flexpack/flexpack.go:25`, and nix, uv and maven all implement it the same way. Returning the checksum struct would be the better design, but it would have to change across every implementation at once, so I'd rather not have RubyGems diverge on its own in this PR.

Happy to raise it separately if you want it done repo-wide.

flexpack/rubygems_flexpack.go
agrasth: Fair point, the hardcoded indent encodes the lock file layout.

This is being removed rather than fixed: per the discussion on line 489, the collector is moving to the dependency graph Bundler reports directly, so the indent handling and the section tracking go away with the parser.

flexpack/rubygems_flexpack.go
agrasth: Agreed, `countLeadingSpaces` is hand rolled.

Like the indent check above it, this helper disappears with the lock file parser.

entities/buildinfo.go
agrasth: This was noted as fixed in an earlier reply, but that commit never actually made it to the branch — it's live now in 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: Done — `rubygems_flexpack.go` now queries Bundler's own `Definition` API (`bundle exec ruby -e ...`) instead of parsing Gemfile.lock. See 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: This helper is removed entirely along with the rest of the lock-file parser in 440b0b6 — dependency names now come straight from Bundler's own graph, no text parsing needed.

flexpack/rubygems_flexpack.go
agrasth: This was marked fixed earlier but the commit never made it to the branch — it's actually renamed and pushed now in 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: Done in 440b0b6 — scopes now come directly from the group names Bundler's own Definition API reports (e.g. `default`, `development`), with a `default` fallback instead of the old hardcoded `production`.

flexpack/rubygems_flexpack.go
agrasth: Same fix as the other two "use a dependency graph command" comments on this file — done in 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: Same fix as the other two "use a dependency graph command" comments on this file — done in 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: This is removed now — the indent-based section tracking went away with the whole lock-file parser in 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: Removed — `countLeadingSpaces` no longer exists as of 440b0b6.

flexpack/rubygems_flexpack.go
agrasth: Moot now — `parseSpecLine` is removed entirely as of 440b0b6 along with the rest of the lock-file parser.

flexpack/rubygems_flexpack.go
agrasth: Still pending — I haven't touched branch history. Everything RubyGems-related is now in 440b0b6; the unrelated-content cleanup (.NET/npm/nix/CI files from stale merge history) is a separate rebase decision we're holding off on for now.

reshmifrog (approved):

PR #395, opened 2026-06-29
RTECO-1401 Fix/Publish Build Info with Build Duration 

reshmifrog:
-----
Problem : the build duration was always coming `0` in the build info.
Solution: added the logic to calculate the build duration and capture it in build info.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -291,7 +291,14 @@ func (b *Build) createBuildInfoFromPartials() (*entities.BuildInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	buildInfo.Started = buildGeneralDetails.Timestamp.Format(entities.TimeFormat)
+	startTime := buildGeneralDetails.Timestamp
+	buildInfo.Started = startTime.Format(entities.TimeFormat)
+	// Record how long the build took from its start until now (build-info generation / publish time).
+	if !startTime.IsZero() {
+		if duration := time.Since(startTime).Milliseconds(); duration > 0 {
+			buildInfo.DurationMillis = duration
+		}
+	}
 	modules, env, vcsList, issues, err := extractBuildInfoData(partials)
 	if err != nil {
 		return nil, err
--- a/build/build_test.go
+++ b/build/build_test.go
@@ -1,10 +1,14 @@
 package build
 
 import (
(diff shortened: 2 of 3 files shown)


build/build.go
```diff
@@ -291,7 +291,14 @@ func (b *Build) createBuildInfoFromPartials() (*entities.BuildInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	buildInfo.Started = buildGeneralDetails.Timestamp.Format(entities.TimeFormat)
+	startTime := buildGeneralDetails.Timestamp
+	buildInfo.Started = startTime.Format(entities.TimeFormat)
+	// Record how long the build took from its start until now (build-info generation / publish time).
+	if !startTime.IsZero() {
+		if duration := time.Since(startTime).Milliseconds(); duration > 0 {
```
agrasth, line 298: `time.Since(startTime)` is the elapsed time from when the build started until `createBuildInfoFromPartials()` runs — which is at publish time, not when the build actually finished. If a user runs jf upload at 10:00, the build finishes at 10:05, but they only run jf bp at 10:30, this will report 30 minutes. The actual build took 5 minutes. 
Is this the intended semantics per RTECO-1401, or should there be an explicit end-time recorded?

build/build.go
```diff
@@ -291,7 +291,14 @@ func (b *Build) createBuildInfoFromPartials() (*entities.BuildInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	buildInfo.Started = buildGeneralDetails.Timestamp.Format(entities.TimeFormat)
+	startTime := buildGeneralDetails.Timestamp
+	buildInfo.Started = startTime.Format(entities.TimeFormat)
+	// Record how long the build took from its start until now (build-info generation / publish time).
```
agrasth, line 296: this comment is not necessary ig

build/build.go
reshmifrog: This is intentional. The CLI has no way to know exactly when the build finished it only knows when the first command ran and when jf rt bp publishes, so it measures the time between those. In the normal case you publish right after building, so the number is accurate, it only looks off if you wait a long time before publishing.

agrasth (approved):

PR #398, opened 2026-07-14
RTECO-945: Add Alpine APK build-info collection support

naveenku-jfrog:
## Summary

Adds native Alpine Linux APK build-info collection to `build-info-go`, enabling `bi apk` to record installed packages as structured Build Info dependencies with checksums, scopes, and `requestedBy` chains — consistent with how other package managers (npm, pip, Maven) are supported in this library.

---

### New: `build/utils/alpine.go`
Core utilities for Alpine package handling:
- **`AlpinePackage` struct** — holds `Name`, `Version`, `Arch`, `Size`, `Origin`, `URL`, `Depends`, `Provides`, and `Files`
- **`ListInstalledPackages()` / `parseInstalledDB()`** — parses `/lib/apk/db/installed` directly (zero subprocesses), returning deduplicated packages with full metadata
- **`DiffAlpinePackages()`** — snapshot-diff model consistent with npm/pip modules in this library
- **`ChecksumsFromCache()`** — computes SHA-1/SHA-256/MD5 by matching a package against a `.apk` archive under a supplied cache directory; retries the lookup briefly to absorb a rare filesystem-visibility lag right after `apk` writes the archive
- **`PackagesFromArchivesDir()`** — parses name/version pairs straight from `.apk` filenames in a directory, recognizing apk's real cache filename format (`<name>-<version>.<8-hex-char-content-hash>.apk`), not just the bare `<name>-<version>.apk` form
- **`BuildDepGraph()` / `parseDependsOutput()`** — builds a forward dependency graph via `apk info --depends` for `requestedBy` chain construction

### New: `build/alpine.go`
The `AlpineModule` that integrates into the build pipeline:
- **`AlpineModule`** — attached to a `Build`, orchestrates pre/post snapshot diffing to identify newly installed packages
- **`SetPreSnapshot()` / `SetCacheDir()` / `SetDownloadsDir()` / `SetRequestedPackages()`** — configure the module before `apk add` runs
- **`CollectDependencies()`** — diffs snapshots, resolves checksums from the cache directory, classifies scope (`prod` vs `transitive`), and builds `requestedBy` chains
- **`depsFromDownloadedArchives()`** — safety net that catches packages apk downloaded into the cache during dependency resolution but that don't show up in the installed-DB diff, so they still get recorded with a checksum
- **`SaveBuildInfo()`** — persists the collected dependencies as an `entities.Alpine` module
- **`buildRequestedBy()`** — populates `RequestedBy` chains with a depth guard (`RequestedByMaxLength = 15`)

### Modified: `build/build.go`, `entities/buildinfo.go`, `cli/cli.go`
- Registered `AlpineModule` in the build factory via `AddAlpineModule()`
- Added `Alpine` package type constant (`entities.Alpine = "alpine"`)
- CLI wiring for the `jf bi apk` command

### New: `build/alpine_test.go`, `build/utils/alpine_test.go`
46 unit tests covering `AlpinePackage.ID()`, `parseInstalledDB()` (including deduplication), `DiffAlpinePackages()` edge cases, `ChecksumsFromCache()` (hit/miss/retry), `PackagesFromArchivesDir()` (including the real hash-suffixed filename format), `BuildDepGraph()` / `parseDependsOutput()`, `buildRequestedBy()`, and `AlpineModule` setters/scope classification/module type.

---

## Design Notes

**Checksum strategy**: checksums come from matching each newly-installed package against a `.apk` archive under a cache directory the caller provides (`SetCacheDir`). This is why the caller (`jf apk add`) creates its own temporary cache directory and passes it to native `apk` via `--cache-dir` — checksums are only available when a cache directory is in play.

**No subprocess dependency for metadata**: Package metadata is parsed directly from `/lib/apk/db/installed` in Go — no runtime dependency on the `apk` binary for the DB read itself.

**Snapshot-diff model**: Consistent with how npm and pip work in this library — snapshot before `apk add`, snapshot after, diff gives the newly installed set. A downloaded-archives fallback additionally catches packages that apk fetched into the cache but that didn't end up as a diff entry.

**Dependency ID format**: `name:version` (e.g. `curl:8.14.1-r2`), consistent with the convention used by npm, pip, and other modules in this library.

---

## Fixes included since the original write-up

- **Archive filename regex was broken for real apk cache output.** `apkArchiveNamePattern` only matched a theoretical `<name>-<version>.apk` filename; real `apk --cache-dir` archives are always named `<name>-<version>.<8-hex-char-content-hash>.apk` (verified against a live Alpine container). This silently made `PackagesFromArchivesDir` — and the whole downloaded-archives safety net — non-functional. Fixed the regex to accept the hash segment (optional, so both forms still match) and added a test case with the real filename format.
- **Added a short retry around the cache-checksum glob lookup**, to absorb a transient filesystem-visibility lag observed in CI where a dependency's checksum came back empty immediately after `apk add` completed, on the first attempt only.


cli/cli.go
```diff
@@ -452,6 +453,52 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
+				cacheDir, err := os.MkdirTemp("", "bi-apk-cache-*")
+				if err != nil {
+					return fmt.Errorf("failed to create apk cache dir: %w", err)
+				}
+				defer func() { _ = os.RemoveAll(cacheDir) }()
+
+				apkArgs = append([]string{"--cache-dir", cacheDir}, apkArgs...)
+
+				alpineModule := bld.AddAlpineModule("apk-module", "", "")
+				alpineModule.SetCacheDir(filepath.Clean(cacheDir))
```
agrasth, line 484: `alpineModule.SetRequestedPackages(...)` is never called here, even though the explicitly-requested package names are already available in `apkArgs` (line 472). Without it, `requested[pkg.Name]` is always false downstream in `buildRequestedBy`/`CollectDependencies`, so every dependency from a `bi apk` run — including the package the user actually asked to install — gets classified as scope `transitive` instead of `prod`. Should this call `alpineModule.SetRequestedPackages(apkArgs)` (or the filtered package-name subset of it) before `SnapshotInstalledPackages`?

build/utils/alpine.go
```diff
@@ -0,0 +1,346 @@
...
+func (p AlpinePackage) SHA1Hex() string {
+	v := strings.TrimPrefix(p.Checksum, "Q1")
+	if v == "" {
+		return ""
+	}
+	raw, err := base64.StdEncoding.DecodeString(v)
+	if err != nil {
+		return ""
+	}
+	return hex.EncodeToString(raw)
```
agrasth: Worth double-checking this against a real Alpine box: the `C:` field in `/lib/apk/db/installed` is documented as the SHA1 of only the package's internal *control* segment, not the full `.apk` archive — so this decoded value likely won't match the artifact's real SHA1 as recorded by Artifactory/Xray for the same file. If that's confirmed, this checksum probably shouldn't be used as the tier-1/primary SHA1 without a note that it's a control-segment hash, not a file hash — or the fallback tiers should always be used for SHA1 too, not just when this is empty.

build/utils/alpine.go
```diff
@@ -0,0 +1,346 @@
...
+	}, nil
+}
+
+// BuildDepGraph runs "apk info -r <name>" for each package and returns a forward dependency
+// graph: pkg.Name → []dependency-name. This is used to compute requestedBy chains.
+// Packages for which apk info is unavailable are simply omitted from the graph.
+func BuildDepGraph(pkgs []AlpinePackage) map[string][]string {
+	graph := make(map[string][]string, len(pkgs))
+	for _, pkg := range pkgs {
+		out, err := exec.Command("apk", "info", "--depends", pkg.Name).Output()
```
agrasth: This forks one `apk info --depends` subprocess per installed package to build the dependency graph, which contradicts the PR's own "no subprocess dependency" design note (which applies to `parseInstalledDB`) and duplicates the dependency list already available in the DB's `D:` field for each package. Could this reuse the already-parsed `Depends` data instead of shelling out again?

build/alpine.go
```diff
@@ -0,0 +1,287 @@
...
+		parentID, parentAdded := idByName[parentName]
+		if !parentAdded {
+			continue
+		}
+		for _, depName := range deps {
+			if _, childAdded := idByName[depName]; !childAdded {
+				continue
+			}
+			// Guard against the RequestedByMaxLength (15) used in other PM integrations.
+			if len(result[depName]) >= 15 {
```
agrasth: Two things here: (1) `15` is a hardcoded literal — `entities.RequestedByMaxLength` is already used for this exact guard in `golang.go:175`, `yarn.go:180`, and `solution.go:138`; worth reusing it here instead of duplicating the magic number. (2) `buildRequestedBy` only ever appends the *direct* parent's ID (`[]string{parentID}`) for one level, whereas the established convention in this repo (e.g. `populateRequestedByField` in golang.go) recursively propagates the *full* ancestor chain via `UpdateRequestedBy`. Is a single-level parent intentional for Alpine, or should deeper transitive chains carry their full lineage like the other modules?

build/alpine.go
```diff
@@ -0,0 +1,287 @@
...
+//
+// Phase 2 — pre-existing dependencies of newly installed packages:
+// Packages that were already on the system before `apk add` but are listed as
+// runtime dependencies (D: field) of newly installed packages are also recorded.
+// Their checksums are resolved from the C: DB field and installed files on disk
+// (no .apk cache is expected for pre-existing packages).
+//
+// Callers may further enrich the returned slice (e.g. with AQL checksums) before
+// calling SaveBuildInfo.
+func (m *AlpineModule) CollectDependencies() ([]entities.Dependency, error) {
```
agrasth, line 64: No test in `alpine_test.go` calls `CollectDependencies()` or `resolveDep()` directly — the core orchestration (checksum-tier fallback, scope classification, wiring into `buildRequestedBy`) is only covered indirectly, if at all. Given this is the main entry point tying the checksum/scope/requestedBy logic together, could use a dedicated test with a fake pre/post snapshot and cache dir.

build/utils/alpine.go
```diff
@@ -0,0 +1,346 @@
...
+	defer func() { _ = f.Close() }()
+
+	var packages []AlpinePackage
+	seen := make(map[string]struct{})
+	var cur AlpinePackage
+	var currentDir string // tracks current F: directory within a stanza
+
+	flush := func() {
+		if cur.Name == "" {
+			return
```
agrasth, line 73: A stanza missing `P:` (name) is silently dropped here with no logging — same for any other malformed/incomplete stanza. Might be worth a `log.Debug`/`log.Warn` so a corrupted or unexpected `installed` DB format doesn't just silently produce a smaller-than-expected package list.

.github/workflows/test.yml
```diff
@@ -98,7 +98,7 @@ jobs:
 
       - name: Tests
         env:
-          GOTOOLCHAIN: 'go1.26.3'
+          GOTOOLCHAIN: 'go1.26.5'
```
reshmifrog: why updating this?

build/utils/alpine.go
```diff
@@ -0,0 +1,252 @@
...
+			children = append(children, child)
+		}
+		if len(children) > 0 {
+			graph[pkg.Name] = children
+		}
+	}
+	return graph
+}
+
+func DiffAlpinePackages(before, after []AlpinePackage) []AlpinePackage {
```
reshmifrog, line 262: why there are no comments above the functions description their use?

build/alpine.go
```diff
@@ -0,0 +1,325 @@
...
+
+import (
+	buildutils "github.com/jfrog/build-info-go/build/utils"
+	"github.com/jfrog/build-info-go/entities"
+	"github.com/jfrog/gofrog/crypto"
+	"github.com/jfrog/gofrog/log"
+)
+
+const (
+	AlpineScopeProd       = "prod"
```
reshmifrog, line 11: we are not supposed to name the scope of the deps explicitly , use whatever native package uses

cli/cli.go
```diff
@@ -452,6 +453,54 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
 		},
+		{
+			Name:      "apk",
+			Usage:     "Generate build-info for an Alpine apk project",
+			UsageText: "bi apk [apk command] [command options]",
+			Flags:     flags,
+			Action: func(context *clitool.Context) (err error) {
+				service := build.NewBuildInfoService()
+				service.SetLogger(logger)
+				bld, err := service.GetOrCreateBuild("apk-build", "1")
```
reshmifrog, line 464: apk-build and 1 what are these values and why are we hard-coding them?

cli/cli.go
```diff
@@ -452,6 +453,54 @@ func GetCommands(logger utils.Log) []*clitool.Command {
...
+					return
+				}
+				defer func() {
+					err = errors.Join(err, bld.Clean())
+				}()
+
+				apkArgs := filterCliFlags(context.Args().Slice(), flags)
+				requestedPkgs := extractApkPackageNames(apkArgs)
+
+				cacheDir, err := os.MkdirTemp("", "bi-apk-cache-*")
```
reshmifrog, line 475: why are we creating cache dir ? it should already be there right?

cli/cli.go
```diff
@@ -631,6 +680,30 @@ func filterCliFlags(allArgs []string, cliFlags []clitool.Flag) []string {
 	return filteredArgs
 }
 
+func extractApkPackageNames(args []string) []string {
+	apkCommands := map[string]struct{}{
```
reshmifrog, line 684: why we maintaining the command names?

reshmifrog (changes requested):

.github/workflows/test.yml
naveenku-jfrog: Addressed it.

build/utils/alpine.go
naveenku-jfrog: Addressed it.

build/alpine.go
naveenku-jfrog: Alpine has no native dependency scopes. Build Info already uses prod / transitive as a convention elsewhere (poetry flexpack, pnpm). npm uses prod/dev because npm itself has that distinction. Downstream Alpine CLI code also expects these exact constants. Inventing Alpine-native scope names would break consistency; leaving scopes empty would lose direct-vs-transitive classification.

cli/cli.go
naveenku-jfrog: Same pattern as every other bi command in cli/cli.go: "npm-build", "pip-build", "poetry-build", "twine-build", etc. These are demo CLI defaults for build-info-go’s own bi binary, not production jf build name/number wiring.

cli/cli.go
naveenku-jfrog: This is intentional. Many Alpine environments (especially Docker) don’t keep /var/cache/apk. The CLI creates a temp dir, passes --cache-dir, then reads .apk archives for checksums. Without that, SHA-256/MD5 often can’t be collected.

cli/cli.go
naveenku-jfrog: Needed so the verb (add, upgrade, …) isn’t treated as a package name. Without skipping command tokens, apk add curl parsing breaks or misclassifies args. The list could maybe be narrowed to install-like commands only, but maintaining a command skip-list is correct.

agrasth (approved):

udaykb2 (approved):

PR #407, opened 2026-07-30
Resolve dynamic project version from uv.lock for UV build info

### LINKED ISSUE #3624: Failed to collect UV build info when pyproject.toml uses dynamic = ["version"]
### Describe the bug

jfrog-cli emits a misleading warning when publishing a Python package with `uv` and a `pyproject.toml` that uses a dynamically resolved version (`dynamic = ["version"]` via `hatch-vcs`). Although the package is successfully published, jfrog-cli fails to collect build info.

The project follows a valid PEP 621 pattern where the version is intentionally not statically defined under [project.version] and is instead resolved dynamically by the build backend.
The warning suggests that JFrog CLI only looks for project.version and does not recognize `project.dynamic = ["version"]` or versions provided by build systems such as hatch-vcs.

### Current behavior

on `jf uv publish`
```bash
15:07:09 [Warn] Failed to collect UV build info: failed to create UV FlexPack collector: failed to load pyproject.toml: project version not found in pyproject.toml (checked [project.version])
```

### Reproduction steps

- make sure to be inside a git repository
- use the uv test project in `testdata/uv/uvproject/pyproject.toml` but replace
  ```toml
  [project]
  version = "0.1.0"
  
  [build-system]
  requires = ["flit_core>=3.2"]
  build-backend = "flit_core.buildapi"
  ```
  with
  ```toml
  [project]
  dynamic = ["version"]
  
  [build-system]
  requires = ["hatchling","hatch-vcs"]
  build-backend = "hatchling.build"

  [tool.hatch.version]
  source = "vcs"

  [tool.hatch.version.raw-options]
  search_parent_directories = true
  ```
- run `jf uv ...` commands to build and publish

### Expected behavior

Build info are published without errors using `jf uv publish`, if pyproject.toml follow a valid PEP 621 pattern

### JFrog CLI version

2.116.0

### Operating system type and version

Linux

### JFrog Artifactory version

_No response_

### JFrog Xray version

_No response_

agrasth:
## Summary

`jf uv publish`/build-info collection failed for projects that use PEP 621's `dynamic = ["version"]` (e.g. via hatch-vcs), since pyproject.toml has no static [project.version] in that case. The UV FlexPack collector required that field unconditionally and errored out immediately, never getting a chance to look anywhere else for the real version.

Fixes jfrog/jfrog-cli#3624.

## Change

- UVPyProjectToml now also parses project.dynamic.
- loadPyProjectToml no longer errors on a missing version when it's declared dynamic — resolution is deferred.
- NewUVFlexPack resolves the version, checked in order:
  1. uv.lock's workspace root package entry, if present (kept as a cheap first check; real uv.lock output never populates this in practice, verified empirically, but harmless to check).
  2. UVConfig.InstalledPackages (already populated from `uv pip list` for sync/install/add/remove) — includes the project's own resolved version once it's installed into the venv.
  3. dist/*.whl and dist/*.tar.gz archives in the working directory (build/publish) — both only reach build-info collection after the build itself already succeeded. `uv version` refuses outright to report a dynamic version under any circumstance (verified: it errors even right after a successful build or install), so there's no CLI command to ask directly — but the build backend already writes the resolved version into the archive's own package metadata (wheel: `*.dist-info/METADATA`; sdist: top-level `PKG-INFO`), the same record that would be published to the index. Reading that directly avoids guessing from the filename: no PEP 427/625 name-escaping to get right, and the archive's declared Name is checked against the project name rather than assumed from a prefix match. Since uv doesn't clean dist/ between builds, multiple versions' archives can coexist; the most recently modified matching one wins.
- If none of the above resolve anything, this no longer fails collector construction (which would throw away dependency/artifact collection too). It logs a warning and falls back to an empty version, producing a "name:" module ID — the same tolerance already in place for PEP 723 inline scripts (see the existing comment in loadPyProjectToml).
- Static-version projects are unaffected — this only changes behavior when `dynamic` includes `"version"`.

## Test plan

- Unit tests in flexpack/tests/unit/uv_flexpack_test.go build real (minimal but valid) wheel/sdist archives via archive/zip and archive/tar+compress/gzip to cover: all three resolution sources, the dist/ freshness tie-break (stale + fresh archives present, newest wins), a corrupt/unreadable archive being skipped rather than breaking resolution, a valid archive for an unrelated package correctly not matching by name, and the empty-version fallback path (no source available at all, lock file with no root entry, dist/ with only an unrelated package).
- `go build ./...` and `go vet ./...` clean.
- `go test ./flexpack/...` and `go test -race ./flexpack/...`: all UV tests pass, including every pre-existing one (no regression).
- End-to-end verification against a real hatch-vcs project (git repo, tagged, pyproject.toml with `dynamic = ["version"]` + `[tool.hatch.version] source = "vcs"`), built with a real jf binary wired to this branch via a local go.mod replace, each time in a freshly created isolated worktree so as not to disturb other in-progress work in the shared checkout:
  - `jf uv build` / `jf uv sync`: no warning, build info correctly records the resolved version as the module ID.
  - Reverted to the original unfixed build-info-go and reran the identical `jf uv build`: reproduced the exact reported warning verbatim, confirming the fix addresses the real bug.
  - Built twice without cleaning dist/ (a stale wheel/sdist pair left next to a freshly built one — 4 files spanning 2 versions): confirmed the module ID picks up the fresh version, not the stale one, by reading the actual embedded METADATA/PKG-INFO of each candidate.
  - Ran `jf uv lock` on a project with no dist/, no venv, and no resolvable lock-file version: confirmed the warning fires and build info is still collected (module ID `name:`) instead of being dropped entirely.
  - Confirmed via `uv version`/`uv version --short --output-format json` (before build, after build, after sync) that uv itself has no command that reports a dynamic version under any circumstance, ruling out a simpler native-CLI approach.
- Ran the full repo test suite (`go test ./...`) on both this branch and unmodified main — the only failures on either are pre-existing, environment-dependent ones in build/utils (npm), build/utils/dotnet/solution (nuget), and flexpack/gradle (local Maven/Ivy), unrelated to this change and identical on both branches.

### DIFF
--- a/flexpack/uv_flexpack.go
+++ b/flexpack/uv_flexpack.go
@@ -147,9 +147,9 @@ func (uf *UVFlexPack) loadPyProjectToml() error {
 		return fmt.Errorf("project name not found in pyproject.toml (checked [project.name])")
 	}
 	// A dynamic version (PEP 621 `dynamic = ["version"]`, e.g. via hatch-vcs) is resolved
-	// later from uv.lock once it's loaded — see resolveDynamicVersion.
+	// later, once uv.lock/dist/installed-packages state is available — see resolveDynamicVersion.
 	if uf.projectVersion == "" && !uf.versionIsDynamic {
-		return fmt.Errorf("project version not found in pyproject.toml (checked [project.version])")
+		return fmt.Errorf("project version not found in pyproject.toml (checked [project.version] and [project.dynamic] for \"version\")")
 	}
 	return nil
 }
--- a/flexpack/uv_flexpack.go
+++ b/flexpack/uv_flexpack.go
@@ -21,20 +21,20 @@ import (
 
 // UVLockFile represents the top-level structure of uv.lock
 type UVLockFile struct {
-	Version  int          `toml:"version"`
-	Revision int          `toml:"revision"`
-	Packages []UVPackage  `toml:"package"`
+	Version  int         `toml:"version"`
+	Revision int         `toml:"revision"`
+	Packages []UVPackage `toml:"package"`
 }
 
 // UVPackage represents a [[package]] entry in uv.lock
 type UVPackage struct {
-	Name            string                         `toml:"name"`
-	Version         string                         `toml:"version"`
-	Source          UVSource                       `toml:"source"`
-	Dependencies    []UVDependencyEdge             `toml:"dependencies"`
-	DevDependencies map[string][]UVDependencyEdge  `toml:"dev-dependencies"`
-	Sdist           *UVArtifact                    `toml:"sdist"`
-	Wheels          []UVArtifact                   `toml:"wheels"`
+	Name            string                        `toml:"name"`
+	Version         string                        `toml:"version"`
+	Source          UVSource                      `toml:"source"`
+	Dependencies    []UVDependencyEdge            `toml:"dependencies"`
+	DevDependencies map[string][]UVDependencyEdge `toml:"dev-dependencies"`
+	Sdist           *UVArtifact                   `toml:"sdist"`
+	Wheels          []UVArtifact                  `toml:"wheels"`
 }
 
 // UVSource is an inline table with exactly one key identifying the source type.
@@ -56,7 +56,7 @@ func (s UVSource) IsWorkspacePackage() bool {
 type UVArtifact struct {
 	URL        string `toml:"url"`
 	Path       string `toml:"path"`
-	Hash       string `toml:"hash"`        // "sha256:<hex>"; absent for git
+	Hash       string `toml:"hash"` // "sha256:<hex>"; absent for git
 	Size       int64  `toml:"size"`
 	UploadTime string `toml:"upload-time"` // ISO 8601; may be absent (revision < 3)
 }
@@ -175,9 +175,10 @@ func isVersionDynamic(dynamic []string) bool {
 //     populate it — cheap to check, correct if present.
 //  2. UVConfig.InstalledPackages (populated from `uv pip list` for sync/install/add/remove),
 //     which includes the project's own resolved version once it's installed into the venv.
-//  3. dist/*.whl or dist/*.tar.gz filenames, which encode the resolved version and are
-//     guaranteed to exist by the time build-info collection runs for build/publish (both
-//     only reach this point after `uv build`/`uv publish` already succeeded).
+//  3. dist/*.whl or dist/*.tar.gz archives — build/publish only reach this point after
+//     `uv build`/`uv publish` already succeeded, and each archive's own embedded package
+//     metadata (wheel: *.dist-info/METADATA; sdist: top-level PKG-INFO) records the version
+//     the backend resolved for it — see resolveVersionFromDist.
 //
 // Returns "" if none of these have anything to offer.
 func (uf *UVFlexPack) resolveDynamicVersion() string {
@@ -187,8 +188,16 @@ func (uf *UVFlexPack) resolveDynamicVersion() string {
 		}
 	}
 	if uf.config.InstalledPackages != nil {
-		if version, ok := uf.config.InstalledPackages[normalizeName(uf.projectName)]; ok && version != "" {
-			return version
+		// Compare via normalizeName on both sides rather than indexing by a normalized key
+		// directly: callers (e.g. jfrog-cli-artifactory's uvInstalledPackages) build this map
+		// with their own lowercase/underscore-only normalization, which doesn't collapse dots
+		// or repeated separators the way full PEP 503 normalization does — a project name
+		// containing "." would silently never match a plain map lookup.
+		want := normalizeName(uf.projectName)
+		for pkgName, pkgVersion := range uf.config.InstalledPackages {
+			if pkgVersion != "" && normalizeName(pkgName) == want {
+				return pkgVersion
+			}
 		}
 	}
 	return uf.resolveVersionFromDist()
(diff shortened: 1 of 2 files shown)


flexpack/uv_flexpack.go
```diff
@@ -122,15 +131,105 @@ func (uf *UVFlexPack) loadPyProjectToml() error {
...
 	uf.projectVersion = uf.pyprojectData.Project.Version
+	uf.versionIsDynamic = isVersionDynamic(uf.pyprojectData.Project.Dynamic)
 	if uf.projectName == "" {
 		return fmt.Errorf("project name not found in pyproject.toml (checked [project.name])")
 	}
-	if uf.projectVersion == "" {
+	// A dynamic version (PEP 621 `dynamic = ["version"]`, e.g. via hatch-vcs) is resolved
+	// later from uv.lock once it's loaded — see resolveDynamicVersion.
+	if uf.projectVersion == "" && !uf.versionIsDynamic {
 		return fmt.Errorf("project version not found in pyproject.toml (checked [project.version])")
```
bhanurp: shouldnt the error be improved now? to project.version and dynamicVersion not found?

flexpack/uv_flexpack.go
```diff
@@ -122,15 +142,182 @@ func (uf *UVFlexPack) loadPyProjectToml() error {
...
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			return "", ""
+		}
+		if err != nil {
+			return "", ""
+		}
+		// PKG-INFO sits directly under the single top-level "{name}-{version}/" directory.
+		if strings.HasSuffix(hdr.Name, "/PKG-INFO") || hdr.Name == "PKG-INFO" {
+			return parsePackageMetadata(tr)
```
naveenku-jfrog: This matches the first entry ending in /PKG-INFO, but an sdist can also contain a nested *.egg-info/PKG-INFO. If tar ordering puts a nested one before the authoritative top-level {name}-{version}/PKG-INFO, we'd read the wrong metadata.

flexpack/uv_flexpack.go
```diff
@@ -122,15 +142,182 @@ func (uf *UVFlexPack) loadPyProjectToml() error {
...
+// workspace root package entry ([[package]] with source.virtual/editable/directory == ".")
+// does NOT carry a version field in practice — verified empirically against real `uv lock`
+// output, which omits it entirely for the root/editable entry regardless of a static or
+// dynamic pyproject.toml version. The real resolved version instead shows up in whichever
+// of these the current uv invocation already produced, checked in order:
+//  1. uv.lock's root package version, in case a future uv release (or another tool) does
+//     populate it — cheap to check, correct if present.
+//  2. UVConfig.InstalledPackages (populated from `uv pip list` for sync/install/add/remove),
+//     which includes the project's own resolved version once it's installed into the venv.
+//  3. dist/*.whl or dist/*.tar.gz filenames, which encode the resolved version and are
```
naveenku-jfrog: Stale after the last commit — this says "filenames, which encode the resolved version", but resolveVersionFromDist now reads the embedded METADATA/PKG-INFO, not the filename (and its own comment says so). Please reword bullet 3 to "reads the resolved version from each archive's embedded metadata" so the two comments don't contradict each other.

flexpack/uv_flexpack.go
```diff
@@ -122,15 +142,182 @@ func (uf *UVFlexPack) loadPyProjectToml() error {
...
+//     only reach this point after `uv build`/`uv publish` already succeeded).
+//
+// Returns "" if none of these have anything to offer.
+func (uf *UVFlexPack) resolveDynamicVersion() string {
+	if uf.lockFileData != nil {
+		if rootPkg := findRootPackage(uf.lockFileData.Packages); rootPkg != nil && rootPkg.Version != "" {
+			return rootPkg.Version
+		}
+	}
+	if uf.config.InstalledPackages != nil {
```
naveenku-jfrog, line 190: Lookup key uses full PEP 503 normalizeName ([-_.]+ → -), but the map is built in jfrog-cli-artifactory's uvInstalledPackages with only ToLower + _→- (dots not normalized, runs not collapsed). For a project name containing . this silently misses (falls through to dist, so no crash)

naveenku-jfrog (approved):
Some minor comments.

flexpack/uv_flexpack.go
agrasth: Valid — fixed in c95f386: PKG-INFO now must be exactly one directory deep (`{top}/PKG-INFO`), so a nested `*.egg-info/PKG-INFO` no longer matches regardless of tar ordering. Added a regression test that reproduces the exact scenario (nested stale entry listed before the real one) and fails without the fix.

flexpack/uv_flexpack.go
agrasth: Valid, thanks — reworded bullet 3 in c95f386 to describe the metadata-based resolution instead of the old filename-based wording.

flexpack/uv_flexpack.go
agrasth: Valid — confirmed uvInstalledPackages() in jfrog-cli-artifactory only does ToLower + `_`->`-`, no dot handling. Fixed in c95f386 by normalizing both sides via a scan instead of indexing by a pre-normalized key, so it's correct regardless of the caller's exact normalization. Added a regression test with a dotted project name that fails without the fix.

bhanurp (approved):

apgrucza:
Can we get this one merged and released soon? Been waiting a while for it.

PR #408, opened 2026-08-03
RTECO-1646 - Add tests for BuildInfo properties and enhance MavenFlexPack functionality

fluxxBot:
-----

- Implement TestAppendProperties to verify the behavior of the Append method in BuildInfo, ensuring it fills gaps without clobbering existing properties and initializes nil target properties.
- Enhance MavenFlexPack with additional fields for module locations and checksum caching to optimize artifact handling.
- Introduce methods for resolving the local Maven repository path and collecting build information from multi-module projects.
- Add comprehensive tests for MavenFlexPack, including dependency graph inversion, module dependency collection, and effective POM parsing.
- Ensure that deployment repository resolution is robust, handling various scenarios for snapshot and release versions.

### DIFF
--- a/entities/buildinfo.go
+++ b/entities/buildinfo.go
@@ -31,6 +31,13 @@ const (
 	BuildInfoEnvPrefix   = "buildInfo.env."
 	RequestedByMaxLength = 15
 
+	// MavenBuildModeProperty records which Maven integration produced the build info; it is stamped
+	// with MavenBuildModeNative by the native (FlexPack) collector and MavenBuildModeLegacy by the
+	// legacy build-info-extractor path, giving both a single comparable marker in the published JSON.
+	MavenBuildModeProperty = BuildInfoEnvPrefix + "JFROG_MAVEN_MODE"
+	MavenBuildModeNative   = "native"
+	MavenBuildModeLegacy   = "legacy"
+
 	// Build type
 	Build ModuleType = "build"
 
@@ -106,6 +113,20 @@ func (targetBuildInfo *BuildInfo) Append(buildInfo *BuildInfo) {
 		}
 		if !exists {
 			targetBuildInfo.Modules = append(targetBuildInfo.Modules, newModule)
+		}
+	}
+	// Carry over properties set by the appended build info (e.g. a build-mode marker recorded by the
+	// collector). Existing values on the target win, so this only fills gaps and never clobbers env
+	// already assembled from partials.
+	if len(buildInfo.Properties) == 0 {
+		return
+	}
+	if targetBuildInfo.Properties == nil {
+		targetBuildInfo.Properties = make(Env)
+	}
+	for key, value := range buildInfo.Properties {
+		if _, exists := targetBuildInfo.Properties[key]; !exists {
+			targetBuildInfo.Properties[key] = value
 		}
 	}
 }
--- a/entities/buildinfo_test.go
+++ b/entities/buildinfo_test.go
@@ -255,6 +255,29 @@ func TestAppendProperties(t *testing.T) {
 		target.Append(&source)
 		assert.Equal(t, MavenBuildModeNative, target.Properties[MavenBuildModeProperty])
 	})
+
+	t.Run("empty string in target is treated as existing", func(t *testing.T) {
+		target := BuildInfo{Properties: Env{"x": ""}}
+		source := BuildInfo{Properties: Env{"x": "v"}}
+		target.Append(&source)
+		assert.Equal(t, "", target.Properties["x"])
+	})
+
+	t.Run("chained appends first writer wins", func(t *testing.T) {
+		target := BuildInfo{}
+		b := BuildInfo{Properties: Env{"key": "b"}}
+		c := BuildInfo{Properties: Env{"key": "c"}}
+		target.Append(&b)
+		target.Append(&c)
+		assert.Equal(t, "b", target.Properties["key"])
+	})
+
+	t.Run("non-nil empty source properties is no-op", func(t *testing.T) {
+		target := BuildInfo{Properties: Env{"keep": "v"}}
+		source := BuildInfo{Properties: Env{}}
+		target.Append(&source)
+		assert.Equal(t, Env{"keep": "v"}, target.Properties)
+	})
 }
 
 func TestToCycloneDxBOM(t *testing.T) {
--- a/entities/buildinfo_test.go
+++ b/entities/buildinfo_test.go
@@ -241,6 +241,22 @@ func TestAppend(t *testing.T) {
 	assert.True(t, results)
 }
 
+func TestAppendProperties(t *testing.T) {
+	t.Run("fills gaps without clobbering", func(t *testing.T) {
+		target := BuildInfo{Properties: Env{"keep": "target", "shared": "target-wins"}}
+		source := BuildInfo{Properties: Env{"shared": "source-loses", "added": "source"}}
+		target.Append(&source)
+		assert.Equal(t, Env{"keep": "target", "shared": "target-wins", "added": "source"}, target.Properties)
+	})
+
+	t.Run("initializes nil target properties", func(t *testing.T) {
+		target := BuildInfo{}
+		source := BuildInfo{Properties: Env{MavenBuildModeProperty: MavenBuildModeNative}}
+		target.Append(&source)
+		assert.Equal(t, MavenBuildModeNative, target.Properties[MavenBuildModeProperty])
+	})
+}
+
 func TestToCycloneDxBOM(t *testing.T) {
 	dependencyA := Dependency{Id: "dependency-a", Checksum: Checksum{Sha1: "dependency-a-sha"}, RequestedBy: [][]string{{"dependency-c"}}}
 	dependencyB := Dependency{Id: "dependency-b", Checksum: Checksum{Sha1: "dependency-b-sha"}, RequestedBy: [][]string{{"dependency-b"}, {"dependency-c"}}}
--- a/flexpack/maven_flexpack.go
+++ b/flexpack/maven_flexpack.go
@@ -1159,11 +1159,12 @@ func (mf *MavenFlexPack) deployURLFromEffectiveSettings(isSnapshot bool) (string
 	}
 	result := ""
 	for _, profile := range settings.Profiles {
-		if isSnapshot && profile.AltSnapshotDeploymentRepository != "" {
+		switch {
+		case isSnapshot && profile.AltSnapshotDeploymentRepository != "":
 			result = repoURLFromAltValue(profile.AltSnapshotDeploymentRepository)
-		} else if !isSnapshot && profile.AltReleaseDeploymentRepository != "" {
+		case !isSnapshot && profile.AltReleaseDeploymentRepository != "":
 			result = repoURLFromAltValue(profile.AltReleaseDeploymentRepository)
-		} else if profile.AltDeploymentRepository != "" {
+		case profile.AltDeploymentRepository != "":
 			result = repoURLFromAltValue(profile.AltDeploymentRepository)
 		}
 	}
--- a/flexpack/maven_flexpack.go
+++ b/flexpack/maven_flexpack.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
@@ -44,6 +45,17 @@ type MavenFlexPack struct {
 // artifactChecksum holds the checksums of a single file, cached across modules by artifact path.
 type artifactChecksum struct {
 	sha1, sha256, md5 string
+}
+
+// checksumCacheKey returns a normalized path suitable for use as a checksumCache map key.
+// On case-insensitive filesystems (macOS default, Windows) the same file can be reached
+// via different-cased paths, so the key is lowercased on those platforms.
+func checksumCacheKey(path string) string {
+	key := filepath.Clean(path)
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+		return strings.ToLower(key)
+	}
+	return key
 }
 
 // ModuleLocation records where a reactor module was built and its packaging, derived from the
@@ -419,11 +431,12 @@ func (mf *MavenFlexPack) calculateChecksumWithFallback(dep DependencyInfo) map[s
 	// Strategy 1: Try to find artifact in Maven local repository. Checksums are cached by path so a
 	// dependency shared across reactor modules is hashed only once.
 	if artifactPath := mf.findMavenArtifact(dep); artifactPath != "" {
-		cached, ok := mf.checksumCache[artifactPath]
+		cacheKey := checksumCacheKey(artifactPath)
+		cached, ok := mf.checksumCache[cacheKey]
 		if !ok {
 			if sha1, sha256, md5, err := mf.calculateFileChecksum(artifactPath); err == nil {
 				cached = artifactChecksum{sha1: sha1, sha256: sha256, md5: md5}
-				mf.checksumCache[artifactPath] = cached
+				mf.checksumCache[cacheKey] = cached
 				ok = true
 			} else {
 				log.Warn(fmt.Sprintf("Failed to calculate checksum for artifact: %s", artifactPath))
@@ -862,9 +875,13 @@ func invertDependencyGraph(graph map[string][]string) map[string][]string {
 // than one parent) yields one path per distinct route. A cycle, which a resolved Maven tree should
 // never contain, is broken defensively by terminating the offending path at the repeated node.
 func buildRequestedByPaths(depID string, parents map[string][]string) [][]string {
+	const maxPaths = 15
 	var paths [][]string
 	var walk func(node string, acc []string, onPath map[string]bool)
 	walk = func(node string, acc []string, onPath map[string]bool) {
+		if len(paths) >= maxPaths {
+			return
+		}
 		directParents := parents[node]
 		if len(directParents) == 0 {
 			// Reached a root: the accumulated ancestors form one complete path.
@@ -874,6 +891,9 @@ func buildRequestedByPaths(depID string, parents map[string][]string) [][]string
 			return
 		}
 		for _, parent := range directParents {
+			if len(paths) >= maxPaths {
+				return
+			}
 			if onPath[parent] {
 				// Cycle guard: close the path at the repeated node instead of recursing forever.
 				paths = append(paths, append(append([]string(nil), acc...), parent))
@@ -885,6 +905,9 @@ func buildRequestedByPaths(depID string, parents map[string][]string) [][]string
 		}
 	}
 	walk(depID, nil, map[string]bool{depID: true})
+	if len(paths) == maxPaths {
+		log.Debug(fmt.Sprintf("dependency %s RequestedBy paths capped at %d", depID, maxPaths))
+	}
 	return paths
 }
 
@@ -1120,7 +1143,7 @@ func distributionURL(project effectivePomProject, fallbackSnapshot bool) string 
 	if dm.Repository.URL != "" {
 		return dm.Repository.URL
 	}
-	return dm.SnapshotRepository.URL
+	return ""
 }
 
 func (mf *MavenFlexPack) deployURLFromEffectiveSettings(isSnapshot bool) (string, error) {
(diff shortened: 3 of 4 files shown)


entities/buildinfo.go
```diff
@@ -108,6 +115,20 @@ func (targetBuildInfo *BuildInfo) Append(buildInfo *BuildInfo) {
...
+	// Carry over properties set by the appended build info (e.g. a build-mode marker recorded by the
+	// collector). Existing values on the target win, so this only fills gaps and never clobbers env
+	// already assembled from partials.
+	if len(buildInfo.Properties) == 0 {
+		return
+	}
+	if targetBuildInfo.Properties == nil {
+		targetBuildInfo.Properties = make(Env)
+	}
+	for key, value := range buildInfo.Properties {
```
agrasth, line 127: **Cross-cutting semantic change on a shared method.** `Append` is used by every builder (Go, npm, Poetry, Ruby, ...), not just Maven. Looks like historically it never touched `Properties`. Now every caller that partially assembles a `BuildInfo` and calls `Append` picks up merged properties, with "first-writer-wins" semantics.

Example: two `Append` calls both contribute `Properties["buildmode"]` — the second is silently dropped. That's easy to misread as a bug.

flexpack/maven_flexpack.go
```diff
@@ -407,6 +460,40 @@ func (mf *MavenFlexPack) calculateChecksumWithFallback(dep DependencyInfo) map[s
...
+	// ExtraArgs nor the home directory changes during a collection.
+	if mf.localRepoResolved {
+		return mf.localRepo
+	}
+	mf.localRepo = mf.resolveLocalRepositoryPath()
+	mf.localRepoResolved = true
+	return mf.localRepo
+}
+
+func (mf *MavenFlexPack) resolveLocalRepositoryPath() string {
```
agrasth, line 490: **Silent correctness bug.** `resolveLocalRepositoryPath` only inspects `ExtraArgs` for `-Dmaven.repo.local=`. It does *not* honor `<localRepository>` in `~/.m2/settings.xml` or `${M2_HOME}/conf/settings.xml`.

Example: a CI runner writes `<localRepository>/opt/m2-cache</localRepository>` into settings.xml (very common). Maven downloads there, FlexPack looks in `~/.m2/repository`, finds nothing → every dependency's checksum comes out empty and the build-info is silently wrong.

Either parse `settings.xml`, or run `mvn help:evaluate -Dexpression=settings.localRepository -q -DforceStdout` once and cache the result.

flexpack/maven_flexpack.go
```diff
@@ -35,8 +60,24 @@ type MavenConfig struct {
...
+// `mvn dependency:tree -DoutputFile=<name>` runs. Maven runs the goal once per module and writes
+// this file into every module's own base directory, so a multi-module build produces one file per
+// module (root aggregator included).
+const mavenDepsFileName = "maven-deps.json"
+
+// mavenDependencyPluginTreeGoal is the fully-qualified plugin coordinate used to invoke the
+// dependency:tree goal. The version is pinned because `-DoutputType=json` was only added in
+// maven-dependency-plugin 3.7.0; older versions silently write plain-text output which the JSON
+// parser then rejects. 3.8.1 (current latest) is used to also pick up bug fixes.
+const mavenDependencyPluginTreeGoal = "org.apache.maven.plugins:maven-dependency-plugin:3.8.1:tree"
```
agrasth, line 91: Pinning `maven-dependency-plugin:3.8.1` forces every collection to resolve that exact plugin coordinate from the configured Maven remote. Air-gapped / mirrored setups that haven't uploaded 3.8.1 now fail with "cannot resolve plugin" instead of using whatever the user already had cached.

Consider: try pinned first, fall back to unversioned `dependency:tree` on resolution failure.

flexpack/maven_flexpack.go
```diff
@@ -690,3 +1013,199 @@ func saveMavenBuildInfoForJfrogCli(buildInfo *entities.BuildInfo) error {
...
+
+// GetDeploymentRepositories resolves where Maven deploys, from its effective model. It returns exactly
+// one of:
+//   - overrideURL: a single URL applied to ALL modules - Maven's -DaltDeploymentRepository or an
+//     active-profile altDeploymentRepository in (effective) settings, which override distributionManagement; or
+//   - moduleURLs: per-module (moduleId "groupId:artifactId:version" -> URL) distributionManagement from
+//     the effective pom, so a reactor whose modules deploy to different repos is handled correctly.
+//
+// Both empty means no deployment repository is configured. Precedence follows Maven (override wins).
+func (mf *MavenFlexPack) GetDeploymentRepositories() (moduleURLs map[string]string, overrideURL string, err error) {
```
agrasth, line 1074: This path can trigger up to **three separate `mvn` invocations** per collection (`dependency:tree`, `help:effective-pom`, `help:effective-settings`). Each Maven cold-start is 10–30s on a large reactor.

Combine into one: `mvn help:effective-pom help:effective-settings dependency:tree -Doutput=... -DoutputFile=...` shares a single JVM/plugin-resolution pass. Real UX difference on big multi-module builds.

flexpack/maven_flexpack.go
```diff
@@ -690,3 +1013,199 @@ func saveMavenBuildInfoForJfrogCli(buildInfo *entities.BuildInfo) error {
...
+	}
+	var settings struct {
+		Profiles []effectiveSettingsProfile `xml:"profiles>profile"`
+	}
+	if err := xml.Unmarshal([]byte(content), &settings); err != nil {
+		return "", fmt.Errorf("failed to parse effective-settings: %w", err)
+	}
+	for _, profile := range settings.Profiles {
+		if isSnapshot && profile.AltSnapshotDeploymentRepository != "" {
+			return repoURLFromAltValue(profile.AltSnapshotDeploymentRepository), nil
```
agrasth: **First-match-wins is wrong for Maven profile precedence.** Maven's actual rule: later active profiles override earlier ones.

Example: user has profile `defaults` (sets `altDeploymentRepository=A`) then profile `ci` (sets `altDeploymentRepository=B`). This loop returns `A`; real Maven deploys to `B`.

Reverse the iteration or track "last non-empty" and return that.

flexpack/maven_flexpack.go
```diff
@@ -690,3 +1013,199 @@ func saveMavenBuildInfoForJfrogCli(buildInfo *entities.BuildInfo) error {
...
+	if project.Version != "" {
+		isSnapshot = strings.HasSuffix(strings.TrimSpace(project.Version), "-SNAPSHOT")
+	}
+	dm := project.DistributionManagement
+	if isSnapshot && dm.SnapshotRepository.URL != "" {
+		return dm.SnapshotRepository.URL
+	}
+	if dm.Repository.URL != "" {
+		return dm.Repository.URL
+	}
```
agrasth, line 1145: `return dm.SnapshotRepository.URL` at the tail is unreachable for a release version *unless* `<repository>` is empty — in which case it silently returns the snapshot URL for a release deploy. That's a silent wrong-target push.

Return `""` here and let the caller decide the fallback.

flexpack/maven_flexpack.go
```diff
@@ -521,48 +817,75 @@ func (mf *MavenFlexPack) CollectBuildInfo(buildName, buildNumber string) (*entit
...
+}
 
-	return buildInfo, nil
+// buildRequestedByPaths returns every ancestor path for depID, walking the child -> parents map from
+// depID's direct parents up to a root (a node that is itself requested by nothing, i.e. the module).
+// Each path lists ancestors nearest-first and ends at the module root, matching the build-info
+// convention and the legacy Maven build-info extractor. A diamond dependency (reached through more
+// than one parent) yields one path per distinct route. A cycle, which a resolved Maven tree should
+// never contain, is broken defensively by terminating the offending path at the repeated node.
+func buildRequestedByPaths(depID string, parents map[string][]string) [][]string {
```
agrasth, line 877: Worst-case path count is O(product-of-parent-counts). A diamond-heavy Maven tree (BOM inheritance in a monorepo) can produce thousands of `RequestedBy` paths for a single dep.

Cap at ~15 paths (matches `RequestedByMaxLength` you already have) and drop the rest with a `log.Debug`.

flexpack/maven_flexpack.go
```diff
@@ -484,31 +568,243 @@ func (mf *MavenFlexPack) validateAndNormalizeScopes(scopes []string) []string {
...
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return nil, fmt.Errorf("mvn dependency:tree failed: %w\nOutput: %s", err, string(output))
+	}
+
+	return mf.findDependencyTreeFiles()
+}
+
+// findDependencyTreeFiles walks the working directory and returns every maven-deps.json file, one per
+// reactor module. WalkDir yields lexical order, keeping module ordering deterministic across runs.
+func (mf *MavenFlexPack) findDependencyTreeFiles() ([]string, error) {
```
agrasth, line 682: `findDependencyTreeFiles` walks the entire working directory. In a polyglot repo (Maven + JS + Python vendored deps) this becomes a lot of stat syscalls.

You already know the reactor from Maven — combine with the earlier `mvn help:evaluate -Dexpression=project.modules` (or the combined invocation suggested above) so paths are handed to you directly.

flexpack/maven_flexpack.go
```diff
@@ -373,16 +416,26 @@ func (mf *MavenFlexPack) calculateChecksumWithFallback(dep DependencyInfo) map[s
...
 
-	// Strategy 1: Try to find artifact in Maven local repository
+	// Strategy 1: Try to find artifact in Maven local repository. Checksums are cached by path so a
+	// dependency shared across reactor modules is hashed only once.
 	if artifactPath := mf.findMavenArtifact(dep); artifactPath != "" {
-		if sha1, sha256, md5, err := mf.calculateFileChecksum(artifactPath); err == nil {
-			checksumMap["sha1"] = sha1
-			checksumMap["sha256"] = sha256
-			checksumMap["md5"] = md5
+		cached, ok := mf.checksumCache[artifactPath]
```
agrasth: `checksumCache` keys by `artifactPath`. On case-insensitive filesystems (macOS default, Windows) the same file can be reached via different-cased paths and cache-miss.

Normalize the key with `filepath.Clean` on all platforms; add `strings.ToLower` on darwin/windows.

entities/buildinfo_test.go
```diff
@@ -241,6 +241,22 @@ func TestAppend(t *testing.T) {
 	assert.True(t, results)
 }
 
+func TestAppendProperties(t *testing.T) {
```
agrasth, line 244: Given how load-bearing the new `Append` semantics are, please add these edge cases:

1. Target `Properties["x"] = ""` (empty string), source has `Properties["x"] = "v"` — is empty considered "exists" (source dropped) or "gap" (source wins)? Current impl: exists wins; not asserted.
2. Chained appends: A.Append(B) then A.Append(C), both set the same key — B wins. Not asserted.
3. Source `Properties` is non-nil but empty (`Env{}`) — early return path at line 121 not exercised.

agrasth (approved):

agrasth (commented):
Added comments

flexpack/maven_flexpack.go
fluxxBot: Fixed. `distributionURL` now returns `""` instead of `dm.SnapshotRepository.URL` for the fallthrough case. A release artifact with no `<repository>` configured will get an empty URL (caller skips it) rather than silently deploying to the snapshot repo.

flexpack/maven_flexpack.go
fluxxBot: Fixed. Changed to a last-match accumulator (track `result`, overwrite on each matching profile, return at end). Maven's later-profile-wins semantics now followed correctly.

flexpack/maven_flexpack.go
fluxxBot: Fixed. Added a `const maxPaths = 15` cap with a `log.Debug` when truncating. Matches the spirit of `RequestedByMaxLength` (which caps the stored path length, not path count).

entities/buildinfo_test.go
fluxxBot: Fixed. Added 3 sub-tests: (1) empty-string target treated as existing (source dropped), (2) chained appends — first writer wins, (3) non-nil empty source is a no-op.

entities/buildinfo.go
fluxxBot: The semantics are intentional and documented in the comment above: "Existing values on the target win, so this only fills gaps and never clobbers env already assembled from partials." In practice `Properties` is only written by FlexPack collectors, not by partial assemblers, so the cross-cutting impact is limited. The 3 new test cases in `TestAppendProperties` (empty-string target, chained appends, empty source) now assert the edge-case behaviour explicitly.

flexpack/maven_flexpack.go
fluxxBot: Fixed. Cache key is now `filepath.Clean(artifactPath)`; on darwin/windows it is also lowercased so the same file reached via different-cased paths hits the cache.

flexpack/maven_flexpack.go
fluxxBot: Valid gap. `-Dmaven.repo.local=` in args is respected but `<localRepository>` in settings.xml is not. Adding full settings.xml parsing is scope-creep for this PR; `help:effective-settings` is already invoked for deploy-URL resolution and its output contains `<localRepository>`, so the right fix is to read it from there when available. Will track as a follow-up.

flexpack/maven_flexpack.go
fluxxBot: Pinning is necessary: `-DoutputType=json` was added in 3.7.0 and older cached versions silently produce plain-text output that the JSON parser then rejects with a confusing error. A fallback to unversioned `dependency:tree` on resolution failure (e.g. air-gapped repo without 3.8.1) is reasonable and will be added as a follow-up.

flexpack/maven_flexpack.go
fluxxBot: Valid optimization. Combining `help:effective-pom`, `help:effective-settings`, and `dependency:tree` into one JVM invocation would cut cold-start overhead significantly on large reactors. The complexity is coordinating different `-Doutput` flags per goal. Will track as a follow-up performance improvement.

flexpack/maven_flexpack.go
fluxxBot: The walk already skips `target/` and dot-directories to avoid statting build output. Using reactor module paths directly (from `mvn help:evaluate -Dexpression=project.modules`) would be cleaner but requires an additional mvn invocation or coupling to the combined-invocation refactor mentioned above. Will revisit alongside that.

flexpack/maven_flexpack.go
```diff
@@ -27,6 +29,41 @@ type MavenFlexPack struct {
...
+	// moduleLocations maps each collected module's id ("groupId:artifactId:version") to where Maven
+	// built it, captured from the dependency-tree output. It is the authoritative module list (it
+	// reflects what Maven actually ran, including profile-activated modules) and lets callers attach
+	// deployed artifacts without re-parsing the pom tree.
+	moduleLocations map[string]ModuleLocation
+	// checksumCache memoizes file checksums by artifact path, so a dependency shared across several
+	// reactor modules is read and hashed only once instead of once per module.
+	checksumCache map[string]artifactChecksum
+	// localRepo caches the resolved Maven local-repository path (see localRepositoryPath).
+	localRepo         string
```
bhanurp, line 41: why is this required?

entities/buildinfo.go
```diff
@@ -31,6 +31,13 @@ const (
 	BuildInfoEnvPrefix   = "buildInfo.env."
 	RequestedByMaxLength = 15
 
+	// MavenBuildModeProperty records which Maven integration produced the build info; it is stamped
+	// with MavenBuildModeNative by the native (FlexPack) collector and MavenBuildModeLegacy by the
+	// legacy build-info-extractor path, giving both a single comparable marker in the published JSON.
+	MavenBuildModeProperty = BuildInfoEnvPrefix + "JFROG_MAVEN_MODE"
```
bhanurp, line 37: where is this getting stamped? what does published JSON mean here is it build-info.json? why do we need to stamp this?

flexpack/maven_flexpack.go
```diff
@@ -484,31 +581,243 @@ func (mf *MavenFlexPack) validateAndNormalizeScopes(scopes []string) []string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	buildInfo := &entities.BuildInfo{
+		Name:    buildName,
+		Number:  buildNumber,
+		Started: time.Now().Format(entities.TimeFormat),
+		Agent: &entities.Agent{
+			Name:    "build-info-go",
```
bhanurp, line 605: agent name is hardcoded and different from previous "GENERIC" and version to match with jfrog-cli version?

flexpack/maven_flexpack.go
```diff
@@ -484,31 +581,243 @@ func (mf *MavenFlexPack) validateAndNormalizeScopes(scopes []string) []string {
...
+			Version: "1.0.0",
+		},
+		BuildAgent: &entities.Agent{
+			Name:    "Maven",
```
bhanurp, line 609: build agent has to be `jfrog-cli-go` and version to match with jfrog-cli version

flexpack/maven_flexpack.go
```diff
@@ -407,6 +473,40 @@ func (mf *MavenFlexPack) calculateChecksumWithFallback(dep DependencyInfo) map[s
...
+	if mf.localRepoResolved {
+		return mf.localRepo
+	}
+	mf.localRepo = mf.resolveLocalRepositoryPath()
+	mf.localRepoResolved = true
+	return mf.localRepo
+}
+
+func (mf *MavenFlexPack) resolveLocalRepositoryPath() string {
+	const flag = "-Dmaven.repo.local="
```
bhanurp:
```suggestion
	const mavenLocalRepoFlag = "-Dmaven.repo.local="
```

flexpack/maven_flexpack.go
```diff
@@ -206,7 +261,7 @@ func (mf *MavenFlexPack) parseWithMavenDependencyTree() error {
 		}
 	}()
 
-	args := []string{"dependency:tree", "-DoutputType=json", "-DoutputFile=maven-deps.json"}
+	args := []string{mavenDependencyPluginTreeGoal, "-DoutputType=json", "-DoutputFile=maven-deps.json"}
```
bhanurp, line 263: why is it changed to org.apache.maven.plugins:maven-dependency-plugin:3.8.1:tree instead of `dependency:tree`? Both are same why is the change now to point to a version?

fluxxBot: We were already using `dependency:tree -DoutputType=json` from the beginning (unversioned). We pinned the plugin coordinate because we ran into cases where setups with an older cached version of the plugin (pre-3.7.0) were silently producing plain-text output instead of JSON, which then broke the parser with a confusing error. Also updated the pin to 3.7.0 — the minimum version that introduced `-DoutputType=json` — rather than 3.8.1.

flexpack/maven_flexpack.go
fluxxBot: The `localRepoResolved` bool is needed because the resolved path can legitimately be an empty string (e.g. if `os.UserHomeDir()` fails), and we can't use `""` to mean "not yet resolved" — we'd re-run the scan on every call. The bool acts as a "computed once" flag so `resolveLocalRepositoryPath` only runs once regardless of the result.

entities/buildinfo.go
fluxxBot: Yeah, it's the `build-info.json`. It gets stamped in two places: the native FlexPack collector sets it in memory in `BuildBuildInfo` (so it's always there for the FlexPack path), and for the legacy extractor path `stampMavenBuildMode` in `jfrog-cli-artifactory` patches it into the JSON file the extractor writes to disk. The reason we need it is to tell which collection path produced the build info — useful for debugging when both paths are active.

flexpack/maven_flexpack.go
fluxxBot: Done, applied.

flexpack/maven_flexpack.go
fluxxBot: The `Agent.Name`/`Agent.Version` fields set here are actually overridden at `jf rt bp` time — `publishBuildInfo` in jfrog-cli calls `SetAgentName("jfrog-cli-go")` and `SetAgentVersion(<cli version>)` before publishing, so whatever we put here is just a fallback if someone uses the library directly. Same for `Agent.Version = "1.0.0"`.\n\nFor `BuildAgent.Name`, we followed the same pattern as Helm, Conan and Nix in this repo (all use the actual build tool name — "Helm", "Conan", "Nix"). `SetBuildAgentName` is never called in the publish path, so "Maven" stays. Happy to change it to "jfrog-cli-go" if that's the preferred convention for Maven specifically, but wanted to flag the inconsistency first.

flexpack/maven_flexpack.go
fluxxBot: The `Agent.Name`/`Agent.Version` fields set here are actually overridden at `jf rt bp` time — `publishBuildInfo` in jfrog-cli calls `SetAgentName("jfrog-cli-go")` and `SetAgentVersion(<cli version>)` before publishing, so whatever we put here is just a fallback if someone uses the library directly. Same for `Agent.Version = "1.0.0"`.\n\nFor `BuildAgent.Name`, we followed the same pattern as Helm, Conan and Nix in this repo (all use the actual build tool name — "Helm", "Conan", "Nix"). `SetBuildAgentName` is never called in the publish path, so "Maven" stays. Happy to change it to "jfrog-cli-go" if that's the preferred convention for Maven specifically, but wanted to flag the inconsistency first.

flexpack/maven_flexpack.go
bhanurp: mention it in comments or leave blank otherwise confusing even for agents to understand what happening here

bhanurp (approved):

flexpack/maven_flexpack.go
fluxxBot: added comments

PR #413, opened 2026-08-13
XRAY-157662 - fix npm peer dependency

Phavya-jfrog:
-----

For unresolved peer dependencies, they should be skipped and reported as missing dependencies. Previously, they were being converted to a SHA1 hash and sent to curation, which returned a 404 for a package version that never existed. Now they are correctly skipped instead of being treated as a resolvable package.

Before:
Screenshot 2026-08-13 at 11 58 47 AM

After:
Screenshot 2026-08-13 at 12 06 07 PM

### DIFF
--- a/build/utils/npm.go
+++ b/build/utils/npm.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/jfrog/gofrog/crypto"
@@ -119,13 +120,23 @@ func CalculateDependenciesMap(executablePath, srcPath, moduleId string, npmListP
 		}
 	}
 	parseFunc := parseNpmLsDependencyFunc(npmVersion)
+	// Dependencies that npm never resolved (e.g. an unmet peer dependency) and that therefore
+	// carry no locator we can hash into a version. They're excluded from dependenciesMap below;
+	// collect their names here so they're still visible to the caller instead of vanishing silently.
+	var unresolvedDeps []string
 	// Parse the dependencies json object.
-	return dependenciesMap, jsonparser.ObjectEach(data, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) (err error) {
+	if err := jsonparser.ObjectEach(data, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) (err error) {
 		if string(key) == "dependencies" {
-			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, log)
+			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, &unresolvedDeps, log)
 		}
 		return err
-	})
+	}); err != nil {
+		return nil, err
+	}
+	if len(unresolvedDeps) > 0 {
+		printMissingDependenciesWarning("unresolved dependency", unresolvedDeps, log)
+	}
+	return dependenciesMap, nil
 }
 
 func runNpmLsWithNodeModules(executablePath, srcPath string, npmArgs []string, log utils.Log) (data []byte) {
@@ -266,11 +277,11 @@ type npmLsDependency struct {
 	InBundle  bool
 	Dev       bool
 	Optional  bool
-	// Missing peer dependency in npm version 7/8
+	// Missing dependency in npm version 7/8 (not peer-specific)
 	Missing bool
-	// Problems with missing peer dependency in npm version 7/8
+	// Problems with missing dependency in npm version 7/8 (not peer-specific)
 	Problems []string
-	// Missing  peer dependency in npm version 6
+	// Missing peer dependency in npm version 6
 	// Bound to 'legacyNpmLsDependency' struct
 	PeerMissing interface{}
 }
@@ -349,10 +360,38 @@ func extractUrlFromProblems(problems []string, packageName string) string {
 			if idx := strings.Index(url, ","); idx > 0 {
 				url = url[:idx]
 			}
-			return strings.TrimSpace(url)
+			url = strings.TrimSpace(url)
+			// npm reports any missing dependency via this same "missing: name@X" shape.
+			if !isNonRegistryLocator(url) {
+				return ""
+			}
+			return url
 		}
 	}
 	return ""
+}
+
+// bareGitHubShorthandWithRef matches npm's un-prefixed "owner/repo#ref" shorthand
+// (e.g. "expressjs/express#v4.18.0"). The ref is required: without one, "owner/repo"
+// has nothing to pin a version to and would only fabricate a hash — same as "github:owner/repo#ref".
+var bareGitHubShorthandWithRef = regexp.MustCompile(`^[\w.-]+/[\w.-]+#.+$`)
+
+// isNonRegistryLocator reports whether s is a fetchable locator (URL, git, file)
+// rather than a semver range or npm alias spec.
+func isNonRegistryLocator(s string) bool {
+	if strings.Contains(s, "://") {
+		return true
+	}
+	lower := strings.ToLower(s)
+	// git+ and git: are listed for readability but never reach this loop in practice —
+	// real npm git specs using either always include "://" right after (git+ssh://,
+	// git+https://, git://), so the check above already catches them first.
+	for _, prefix := range []string{"git+", "git:", "file:", "github:", "gitlab:", "bitbucket:", "gist:"} {
+		if strings.HasPrefix(lower, prefix) {
+			return true
+		}
+	}
+	return bareGitHubShorthandWithRef.MatchString(s)
 }
 
 func (nld *npmLsDependency) getScopes() (scopes []string) {
@@ -371,7 +410,9 @@ func (nld *npmLsDependency) getScopes() (scopes []string) {
 }
 
 // Parses npm dependencies recursively and adds the collected dependencies to the given dependencies map.
-func parseDependencies(data []byte, pathToRoot []string, dependencies map[string]*dependencyInfo, parseFunc func(data []byte) (*npmLsDependency, error), log utils.Log) error {
+// Dependencies that npm never resolved (missing, with no locator to hash into a version) are excluded
+// from dependencies and their names appended to unresolvedDeps instead, so the caller can still report them.
+func parseDependencies(data []byte, pathToRoot []string, dependencies map[string]*dependencyInfo, parseFunc func(data []byte) (*npmLsDependency, error), unresolvedDeps *[]string, log utils.Log) error {
 	return jsonparser.ObjectEach(data, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
 		if string(value) == "{}" {
 			// Skip missing optional dependency.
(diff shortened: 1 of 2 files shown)


build/utils/npm.go
```diff
@@ -349,12 +359,32 @@ func extractUrlFromProblems(problems []string, packageName string) string {
...
 }
 
+// isNonRegistryLocator reports whether s is a fetchable locator (URL, git, file)
+// rather than a semver range or npm alias spec.
+func isNonRegistryLocator(s string) bool {
+	if strings.Contains(s, "://") {
+		return true
+	}
+	lower := strings.ToLower(s)
+	for _, prefix := range []string{"git+", "git:", "file:", "github:", "gitlab:", "bitbucket:", "gist:"} {
```
gauriy-tech, line 389: **Medium — missed locator: bare `owner/repo` GitHub shorthand**

`isNonRegistryLocator` classifies a locator by checking for `"://"` or one of `git+`, `git:`, `file:`, `github:`, `gitlab:`, `bitbucket:`, `gist:` prefixes. npm also supports specifying a git dependency as a bare `<githubname>/<githubrepo>` (no `github:` prefix, no `://`), e.g. `"dep": "expressjs/express"` in `package.json`. When such a dependency is unresolved, npm's problem string is `missing: dep@expressjs/express, required by root`. Since that string has neither `"://"` nor a recognized prefix, it's misclassified as a semver range and the dependency is silently dropped into `unresolvedDeps` instead of being resolved — a real, fetchable dependency vanishes from build-info entirely.

Verified with a failing test against current code:

```go
func TestIsNonRegistryLocator_GitHubShorthand(t *testing.T) {
	assert.True(t, isNonRegistryLocator("expressjs/express"),
		"GitHub shorthand 'user/repo' is a documented npm git-dependency form and should be treated as a fetchable locator")
}
```

Suggest extending `isNonRegistryLocator` to also match the bare `owner/repo` shorthand (e.g. a regex like `^[\w.-]+/[\w.-]+(#.*)?$`), or documenting why it's intentionally out of scope. No false-positive risk found for scoped packages (`@scope/name`) — the scope slash is already consumed by the `"missing: name@"` prefix strip before this function ever sees the value.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm.go
```diff
@@ -119,13 +119,23 @@ func CalculateDependenciesMap(executablePath, srcPath, moduleId string, npmListP
...
-			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, log)
+			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, &unresolvedDeps, log)
 		}
 		return err
-	})
+	}); err != nil {
+		return nil, err
+	}
+	if len(unresolvedDeps) > 0 {
+		printMissingDependenciesWarning("unresolved dependency, such as an unmet peer dependency", unresolvedDeps, log)
```
gauriy-tech: **Medium — `unresolvedDeps` can contain duplicate names**

`parseDependencies` appends `npmLsDependency.Name` to `unresolvedDeps` on every visit, with no de-duplication. `printMissingDependenciesWarning` (unchanged, `build/utils/npm.go:715-717`) renders the list via a plain `strings.Join`, so a single package that's an unmet dependency of several sibling packages — a common real shape, not an edge case — will appear multiple times in the resulting log line, e.g. `"...skip this dependency: react,react,react-dom"`. A support engineer reading only this log would reasonably read that as 3 distinct problems instead of 1.

Verified with a failing test against current code:

```go
func TestParseDependencies_UnresolvedDepsDeduplicated(t *testing.T) {
	inputJson := `{
		"pkg-a": {"version": "1.0.0", "dependencies": {
			"react": {"problems": ["missing: react@^18.0.0, required by pkg-a@1.0.0"]}
		}},
		"pkg-b": {"version": "1.0.0", "dependencies": {
			"react": {"problems": ["missing: react@^18.0.0, required by pkg-b@1.0.0"]}
		}}
	}`
	depsMap := make(map[string]*dependencyInfo)
	var unresolvedDeps []string
	err := parseDependencies([]byte(inputJson), []string{"root"}, depsMap, npmLsDependencyParser, &unresolvedDeps, &utils.NullLog{})
	assert.NoError(t, err)
	assert.Equal(t, []string{"react"}, unresolvedDeps,
		"expected 'react' to be reported once even though it is missing under two different parents")
}
```

Suggest de-duplicating before appending, or once here before the `printMissingDependenciesWarning` call.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm.go
```diff
@@ -119,13 +119,23 @@ func CalculateDependenciesMap(executablePath, srcPath, moduleId string, npmListP
...
 		if string(key) == "dependencies" {
-			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, log)
+			err = parseDependencies(value, []string{moduleId}, dependenciesMap, parseFunc, &unresolvedDeps, log)
 		}
 		return err
-	})
+	}); err != nil {
+		return nil, err
+	}
+	if len(unresolvedDeps) > 0 {
```
gauriy-tech, line 136: **Low (informational) — a resolved dependency can also appear in the "unresolved" warning**

`unresolvedDeps` tracks by name only and is never cross-checked against the final `dependenciesMap` before `printMissingDependenciesWarning` is called. So a package resolved correctly under one branch (e.g. `react:17.0.2` under sibling `pkg-a`) can simultaneously show up in the warning because a *different* sibling parent (`pkg-b`) has an unmet peer requirement on the same name at a different version. Reproduced directly against the real code — `dependenciesMap` correctly contains the resolved entry while `unresolvedDeps` simultaneously contains the bare name from the sibling branch.

This doesn't corrupt build-info output (the resolved dependency still lands correctly in the final list, and `unresolvedDeps` is a local variable never exposed on the exported API — log-only effect, and logged at `Debug` level), so it's not required for merge. Worth noting: a cheap fix for this would also cover the duplicate-name issue above — filtering `unresolvedDeps` against `dependenciesMap`'s names right before the warning call (a ~10-line pass, since both are fully populated by that point) dedupes and drops any name that turned out to be resolved elsewhere, in one shot:

```go
func filterTrulyUnresolved(dependenciesMap map[string]*dependencyInfo, unresolvedDeps []string) []string {
	resolvedNames := make(map[string]bool, len(dependenciesMap))
	for _, dep := range dependenciesMap {
		resolvedNames[dep.Name] = true
	}
	seen := make(map[string]bool, len(unresolvedDeps))
	var trulyUnresolved []string
	for _, name := range unresolvedDeps {
		if resolvedNames[name] || seen[name] {
			continue
		}
		seen[name] = true
		trulyUnresolved = append(trulyUnresolved, name)
	}
	return trulyUnresolved
}
```

Optional — flagging for awareness / a possible follow-up, not blocking.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm.go
gauriy-tech, line 389: **Low — missing direct test coverage for some locator prefixes; two prefixes are unreachable**

`isNonRegistryLocator` defines 7 explicit prefixes (`git+`, `git:`, `file:`, `github:`, `gitlab:`, `bitbucket:`, `gist:`), but no test in `npm_test.go` directly (or indirectly) exercises the `github:`, `gitlab:`, `bitbucket:`, or `gist:` branches. Separately, `git+` and `git:` are only ever reachable in practice alongside a `"://"` (e.g. `git+ssh://`, `git://`), which the earlier `strings.Contains(s, "://")` check already catches first — so those two prefix entries are effectively unreachable dead branches given real npm syntax.

Suggest adding table cases for the untested prefixes (or a dedicated `TestIsNonRegistryLocator`), and considering removing/commenting on the unreachable `git+`/`git:` entries for clarity. Not blocking — a coverage gap, not a behavioral defect.

---
🤖 Posted by automated PR review (xray-pr-review)

Phavya-jfrog: 1. Treat `expressjs/express` as a locator (as suggested)
2. No `#ref` → falls back to hashing the raw string
3. `sha1("expressjs/express")` = `1f95ce3cf0e43f9a7b65947ce25bd3ef05c408fa`
4. That hash gets recorded as express's "version"
5. Curation checks it against the registry → `express-1f95ce3cf0e43f9a7b65947ce25bd3ef05c408fa.tgz` → 404, same as `react-a25a783e....tgz` did for the original ticket bug

Tried this manually against a real curation instance (not just the code trace) and got exactly that 404.

build/utils/npm.go
Phavya-jfrog: Confirmed, reproduced it — accurate. Leaving as-is for this PR per your own call that it's optional/non-blocking: log-only, Debug level, doesn't touch dependenciesMap or the exported API, and doesn't occur in the scenario this PR actually fixes.

build/utils/npm.go
gauriy-tech: Verified independently — traced the code path (`extractUrlFromProblems` → `extractVersionFromGitUrl` → SHA1-hash fallback in `parseDependencies`) and reproduced the exact hash myself: `sha1("expressjs/express") = 1f95ce3cf0e43f9a7b65947ce25bd3ef05c408fa`. Also confirmed on `main` that pre-PR there was no locator gate at all — any string after `name@` was hashed unconditionally, which is the original ticket bug.

Agreed: extending the locator gate to accept a bare `owner/repo` with no `#ref` would reintroduce a variant of that exact bug, since there's no ref for `extractVersionFromGitUrl` to pull a real version from. Withdrawing this finding — current behavior (drop it, report via `unresolvedDeps`) is correct.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm.go
```diff
@@ -349,12 +359,35 @@ func extractUrlFromProblems(problems []string, packageName string) string {
...
+			}
+			return url
 		}
 	}
 	return ""
 }
 
+// isNonRegistryLocator reports whether s is a fetchable locator (URL, git, file)
+// rather than a semver range or npm alias spec.
+func isNonRegistryLocator(s string) bool {
```
gauriy-tech, line 381: **Low (optional) — `owner/repo#ref` is still conservatively dropped even though it has a resolvable version**

Bare `owner/repo` with no ref is correctly left unresolved (see the withdrawn Finding #1 discussion above — hashing it would reintroduce the original bug). But `owner/repo#ref` *with* an explicit ref is different: it has a real, non-hashed version available via `extractVersionFromGitUrl`, exactly like the already-accepted `github:owner/repo#ref` form does (confirmed: `github:expressjs/express#v4.18.0` resolves to version `v4.18.0`, not a hash). Currently the bare (unprefixed) `#ref` form still gets dropped as unresolved, since it matches neither `"://"` nor any recognized prefix.

Not a bug — just slightly more conservative than it needs to be. Not blocking; a nice-to-have if you want npm's implicit-github-shorthand-with-ref case to resolve like its `github:`-prefixed equivalent does.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm.go
gauriy-tech: **Low (optional) — resulting log message reads awkwardly**

`printMissingDependenciesWarning`'s template is designed for a short noun-phrase "type" (the other call sites pass `"peerDependency"`, `"bundleDependencies"`, `"optionalDependencies"`). This call site passes a full descriptive clause instead, and the rendered result is a grammatically awkward run-on:

> "...because the package is a 'unresolved dependency, such as an unmet peer dependency', which was not manually installed. It is therefore okay to skip this dependency: ..."

("a 'unresolved..." should be "an", and the sentence reads oddly with a full clause dropped into a noun slot.)

Suggest a short label like `"unresolved dependency"` for the templated slot, and keep the "e.g. unmet peer dependency" detail in the inline code comment (already present above, at the `unresolvedDeps` declaration). Not blocking — cosmetic log clarity only.

---
🤖 Posted by automated PR review (xray-pr-review)

build/utils/npm_test.go
```diff
@@ -663,3 +724,29 @@ func TestParseDependenciesEdgeCases(t *testing.T) {
 		})
 	}
 }
+
+func TestNpmIsNonRegistryLocator(t *testing.T) {
```
gauriy-tech, line 742: **Low (optional) — no test locks in the bare `owner/repo` decision**

This test table (`TestNpmIsNonRegistryLocator`) is a great addition, but given how much back-and-forth this thread went through over bare `owner/repo` (see the withdrawn Finding #1 above), it's missing a negative case asserting `isNonRegistryLocator("expressjs/express") == false`. The current table only covers the *prefixed* `github:`/`gitlab:`/`bitbucket:`/`gist:` forms.

Adding that one case would guard against a future contributor "fixing" this again without re-deriving the reasoning from this review thread. Not blocking.

---
🤖 Posted by automated PR review (xray-pr-review)

gauriy-tech (approved):

reshmifrog (approved):

PR #415, opened 2026-08-24
RTECO-1574 - Implementation of NuGet FlexPack support

bhanurp:
## Summary

- Implement NuGet FlexPack build-info collection for `jf nuget` and `jf dotnet` commands
- Support all NuGet project types: SDK-style `.csproj`, legacy `.csproj`/`.fsproj` with `packages.config`, `.sln` multi-project, and standalone `packages.config`
- Support `.snupkg` (modern symbol packages via `symbolpackage/` endpoint) and `.symbols.nupkg` (legacy, flat at repo root)
- Fix artifact `Path` derivation: `.snupkg` → `symbolpackage/<id>.<version>.nupkg`; `.symbols.nupkg` → flat `<id>.<version>.nupkg`; `.nupkg` → flat `<id>.<version>.nupkg`
- Add `BuildArtifactModules` to group pushed/packed artifacts into per-package modules
- Add `CollectPushArtifacts`, `CollectPackedArtifacts`, `SnapshotPackageFiles` helpers
- Support `.slnx` solution format

## Test plan

- [ ] `TestFindNupkgArtifactsIncludesSymbols` — verifies artifact path derivation for all three file types
- [ ] `TestCollectPushArtifactsUsesOnlyExplicitArguments` — verifies only explicit push args are captured
- [ ] `TestCollectPackedArtifactsFindsNestedAndCustomOutputs` — verifies snapshot-diff detection
- [ ] `TestBuildArtifactModules` — verifies module grouping by package ID:version and override


build/utils/dotnet/solution/project/project.go
```diff
@@ -2,22 +2,90 @@ package project
...
 	MarshalJSON() ([]byte, error)
 	Extractor() dependencies.Extractor
 	CreateDependencyTree(log utils.Log) error
 	Load(dependenciesSource string, log utils.Log) (Project, error)
 }
 
-func CreateProject(name, rootPath string) Project {
-	return &project{name: name, rootPath: rootPath}
+// CreateProject creates a Project for the given project file (.csproj/.fsproj/.vbproj/...).
+func CreateProject(projectFilePath string) Project {
```
agrasth, line 31: **Breaking API change on an exported constructor.** `CreateProject(name, rootPath string) → CreateProject(projectFilePath string)`. 
Any external caller doing `project.CreateProject("Foo", "/path")` won't compile. 
We. need to be carefull if every callsite in `jfrog-cli-artifactory` (PR [#532](https://github.com/jfrog/jfrog-cli-artifactory/pull/532)) and `jfrog-cli` (PR [#3664](https://github.com/jfrog/jfrog-cli/pull/3664)) is updated in the same release window, and note this in the release changelog.

build/utils/dotnet/solution/project/project.go
```diff
@@ -2,22 +2,90 @@ package project
...
+}
+
+type msbuildProject struct {
+	PropertyGroups []msbuildPropertyGroup `xml:"PropertyGroup"`
+}
+
+// resolvePackageID reads projectFilePath and returns its effective PackageId, following
+// MSBuild's own default resolution order (PackageId, then AssemblyName), falling back to
+// fallbackName when the file can't be read/parsed or neither property is set.
+func resolvePackageID(projectFilePath, fallbackName string) string {
```
agrasth, line 73: `resolvePackageID` only reads the raw `.csproj`. Real .NET projects commonly declare `<PackageId>` in an imported `Directory.Build.props`; that gets silently missed here and we fall back to `AssemblyName` / filename. 

Example: `.csproj` has no `<PackageId>` but a sibling `Directory.Build.props` says `<PackageId>Company.Real</PackageId>` → this returns `"SampleLib"` (from filename), pack produces `Company.Real.1.0.0.nupkg`, and the build-info tree splits into two modules — exactly the mismatch this PR is trying to fix. 
Either we can walk up the tree for `Directory.Build.props`, or document the limitation in the doc-comment.

build/utils/dotnet/dependencies/dependencies.go
```diff
@@ -27,6 +27,10 @@ type Extractor interface {
 	DirectDependencies() ([]string, error)
 	// Dependencies relations map
 	ChildrenMap() (map[string][]string, error)
+	// ProjectVersion returns the project's own package version when the dependency source
+	// exposes it (project.assets.json), or an empty string when it is unavailable
+	// (e.g. legacy packages.config projects).
+	ProjectVersion() string
```
agrasth, line 33: Is this change intentional?
It can be a **breaking interface change.** Adding `ProjectVersion() string` to the exported `Extractor` interface means anything outside this repo that implements `Extractor` (mocks, third-party extractors) fails to compile.

build/utils/dotnet/solution/solution.go
```diff
@@ -20,6 +20,11 @@ import (
 
 type Solution interface {
 	BuildInfo(module string, log utils.Log) (*buildinfo.BuildInfo, error)
+	// BuildInfoWithNameVersionModuleId behaves like BuildInfo but, when no module override is
+	// given, defaults each module ID to the fixed "<Name>:<Version>" form using the project's
+	// own version (from project.assets.json). Used by the FlexPack path; BuildInfo retains the
+	// legacy project-name default to avoid changing existing 'jf rt nuget/dotnet' build-info.
+	BuildInfoWithNameVersionModuleId(module string, log utils.Log) (*buildinfo.BuildInfo, error)
```
agrasth, line 27: This can be also a **breaking interface change.** 
Same as `Extractor.ProjectVersion()` — adding `BuildInfoWithNameVersionModuleId(...)` to the exported `Solution` interface breaks any external implementer. Please flag in release notes.

build/utils/dotnet/solution/solution.go
```diff
@@ -274,9 +376,22 @@ func (solution *solution) loadSingleProjectFromDir(log utils.Log) error {
...
+		// which is always true here since this directory's name is a path-prefix component of
+		// everything nested under it (e.g. testdata fixtures for unrelated projects).
+		lowerDir := strings.ToLower(solution.path)
+		for _, source := range solution.dependenciesSources {
+			if strings.ToLower(filepath.Dir(source)) == lowerDir {
+				return solution.loadProjectWithSource(project.CreateProjectFromDir(solution.path), source, log)
+			}
+		}
 	}
 	log.Warn(fmt.Sprintf("expecting 1 'proj' file but fuond %d files in path: %s", len(projFiles), solution.path))
```
agrasth: Pre-existing typo `fuond` → `found`. Since you're touching this file, worth fixing in the same PR.

entities/buildinfo.go
```diff
@@ -275,11 +275,14 @@ func mergeArtifacts(mergeArtifacts *[]Artifact, intoArtifacts *[]Artifact) {
...
-		// PRIORITY 1: Check SHA1 - if checksums match, prefer the artifact with a real path.
+		// PRIORITY 1: Check name and SHA1 - if both match, prefer the artifact with a real path.
 		// path="." means the artifact was recorded locally before upload; a non-"." path
 		// means it was confirmed in Artifactory. Always keep the richer entry.
+		// Name must match too - otherwise two distinct artifacts that happen to share
+		// content (e.g. a .nupkg and its .snupkg symbol package) would collide and one
+		// would be silently dropped instead of both being kept.
 		for i, existingArtifact := range *intoArtifacts {
-			if newArtifact.Sha1 == existingArtifact.Sha1 {
+			if newArtifact.Name == existingArtifact.Name && newArtifact.Sha1 == existingArtifact.Sha1 {
```
agrasth, line 285: **Shared-helper behavior change.** `mergeArtifacts` now dedups by `Name && Sha1` instead of `Sha1` alone. Correct for NuGet's `.nupkg` vs `.snupkg` case (which is why you made the change), but this helper is called by every builder — anyone who previously uploaded the same binary twice under a different name got a single entry; now they get two. 
Worth a doc update or release-note bullet so downstream builders aren't surprised.

flexpack/nuget/nuget_flexpack.go
```diff
@@ -0,0 +1,128 @@
...
+	targetPath := n.config.TargetPath
+	if targetPath != "" && !filepath.IsAbs(targetPath) {
+		targetPath = filepath.Join(workingDirectory, targetPath)
+	}
+
+	var (
+		sol buildinfosolution.Solution
+		err error
+	)
+	if strings.HasSuffix(strings.ToLower(filepath.Ext(targetPath)), "proj") {
```
agrasth: `strings.HasSuffix(strings.ToLower(filepath.Ext(targetPath)), "proj")` also matches something like `foo.notaproj`. 
We can use an explicit set (`.csproj`, `.fsproj`, `.vbproj`, `.proj`) or at least check `strings.HasPrefix(ext, ".")`. 
Very low risk in practice but easy to tighten.

flexpack/nuget/nuget_flexpack.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("collect NuGet build info: %w", err)
+	}
+	bi.Name = buildName
+	bi.Number = buildNumber
+	return bi, nil
+}
+
+// GetProjectDependencies returns the flat list of dependencies for the project.
+func (n *NuGetFlexPack) GetProjectDependencies() ([]buildinfoflex.DependencyInfo, error) {
```
agrasth, line 83: `GetProjectDependencies` calls `CollectBuildInfo("", "")` from scratch. 
If a caller also calls `GetDependencyGraph` (line 99) and the real `CollectBuildInfo`, the solution is parsed three times. 
Cache the parsed `*BuildInfo` on the receiver on first call.

flexpack/nuget/nuget_flexpack.go
```diff
@@ -0,0 +1,128 @@
...
+	bi, err := n.CollectBuildInfo("", "")
+	if err != nil {
+		return nil, err
+	}
+	graph := make(map[string][]string)
+	for _, mod := range bi.Modules {
+		for _, d := range mod.Dependencies {
+			for _, chain := range d.RequestedBy {
+				if len(chain) > 0 && chain[0] != "" {
+					graph[chain[0]] = append(graph[chain[0]], d.Id)
```
agrasth: `graph[chain[0]] = append(...)` doesn't dedup edges. 
If the same dep is reached via multiple `RequestedBy` chains sharing `chain[0]`, the child appears twice in the parent's list. 
Cheap fix — use a `map[string]struct{}` and materialize.

flexpack/nuget/nuget_artifacts.go
```diff
@@ -0,0 +1,285 @@
...
+// depending on a single directory and without including pre-existing, unrelated packages.
+func CollectPackedArtifacts(root string, before PackageSnapshot, repoName string) ([]entities.Artifact, error) {
+	after, err := SnapshotPackageFiles(root)
+	if err != nil {
+		return nil, err
+	}
+	var artifacts []entities.Artifact
+	for path, modTime := range after {
+		prev, existed := before[path]
+		if existed && !modTime.After(prev) {
```
agrasth: `modTime.After(prev)` misses a re-pack that lands within the filesystem's mtime resolution (FAT32 = 1 second, some SMB shares are worse). 
If someone runs `pack` twice in quick succession, the second output won't be recorded. 
We can add a Sha1 fallback tie-breaker. WDYT?

flexpack/nuget/nuget_artifacts.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// PackageSnapshot records the package files present under a directory tree, keyed by
+// absolute path, with their last-modified time. It is used to detect the packages that a
+// pack command produces or refreshes.
+type PackageSnapshot map[string]time.Time
+
+// SnapshotPackageFiles walks root recursively and records every existing NuGet package file
+// with its modification time. Missing roots yield an empty snapshot rather than an error.
+func SnapshotPackageFiles(root string) (PackageSnapshot, error) {
```
agrasth: `SnapshotPackageFiles(root)` walks the whole tree — real IO in a monorepo. Bound to `bin/`, `obj/`, `artifacts/`, and the explicit `--output` arg if one was passed.

build/utils/dotnet/solution/project/project.go
bhanurp: Fixed — added to the  section in RELEASE.md as a breaking change.

build/utils/dotnet/solution/project/project.go
bhanurp: Fixed added a doc-comment to `resolvePackageID` explicitly noting that `Directory.Build.props` files in parent directories are not walked, and that the filename fallback is used when `PackageId` is only declared there.

build/utils/dotnet/dependencies/dependencies.go
bhanurp: Yes, intentional `ProjectVersion()` is needed so callers can read the resolved package version without re-parsing. Added to RELEASE.md as a breaking interface change.

build/utils/dotnet/solution/solution.go
bhanurp: Fixed — noted in RELEASE.md as a breaking interface change alongside the `Extractor.ProjectVersion()` entry.

build/utils/dotnet/solution/solution.go
bhanurp: Fixed.

entities/buildinfo.go
bhanurp: Fixed added a release-note bullet in RELEASE.md documenting that `mergeArtifacts` now deduplicates by both `Name` and `Sha1` instead of `Sha1` alone, and explaining the reason.

flexpack/nuget/nuget_flexpack.go
bhanurp: Fixed replaced `strings.HasSuffix(..., "proj")` with an explicit extension set (`.csproj`, `.fsproj`, `.vbproj`, `.proj`).

flexpack/nuget/nuget_flexpack.go
bhanurp: Added a TODO comment noting the triple-parse and that caching the parsed result on the receiver is a future optimization. Left as-is for now to keep the PR scope contained.

flexpack/nuget/nuget_flexpack.go
bhanurp: Fixed `GetDependencyGraph` now deduplicates edges using a `map[string]struct{}` per parent before materializing the slice.

flexpack/nuget/nuget_artifacts.go
bhanurp: Fixed `PackageSnapshot` now stores both `ModTime` and `Sha1` per file. `CollectPackedArtifacts` skips a file only when mtime is not newer **and** SHA1 matches, so a re-pack within FAT32/SMB mtime resolution (1 second) is still detected via content change.

flexpack/nuget/nuget_artifacts.go
bhanurp: Fixed `SnapshotPackageFiles` now only walks `bin/`, `obj/`, `artifacts/`, and any explicit `--output` directory passed via `extraDirs`. Missing directories yield an empty snapshot rather than an error.

flexpack/nuget/nuget_artifacts.go
```diff
@@ -0,0 +1,364 @@
...
+// relative to the "before" snapshot. It scans the same set of directories as the "before"
+// snapshot: the standard output directories (bin/, obj/, artifacts/) under root plus any
+// extraDirs (e.g. a custom --output directory passed to the pack command).
+func CollectPackedArtifacts(root string, before PackageSnapshot, repoName string, extraDirs ...string) ([]entities.Artifact, error) {
+	after, err := SnapshotPackageFiles(root, extraDirs...)
+	if err != nil {
+		return nil, err
+	}
+	var artifacts []entities.Artifact
+	for path, entry := range after {
```
fluxxBot, line 329: for path, entry := range after iterates a map[string]snapshotEntry, so CollectPackedArtifacts returns artifacts in random order per run. Tests work around it by sorting names before comparing. Build-info consumers that depend on stable ordering will see nondeterminism. Collect paths into a sorted []string slice before iterating.

flexpack/flexpack.go
```diff
@@ -160,6 +160,21 @@ type GradleConfig struct {
...
 
+// NuGetConfig holds configuration for the NuGet FlexPack implementation.
+type NuGetConfig struct {
+	WorkingDirectory string
+	// TargetPath is the optional solution, project, or directory passed to the native restore command.
+	// Relative paths are resolved from WorkingDirectory.
+	TargetPath               string
+	ToolchainType            int
+	UseNugetV2               bool
+	AllowInsecureConnections bool
```
fluxxBot: ToolchainType int, UseNugetV2 bool, AllowInsecureConnections bool are declared but never read in this PR. No constants define valid ToolchainType values. Either wire them up, document them with // reserved for caller, or remove until they're needed — otherwise callers have no idea what to pass.

fluxxBot (approved):

agrasth (approved):

PR #416, opened 2026-08-24
RTECO-1885 - Fix Gradle -D/-P property values getting wrapped in literal quotes

naveenku-jfrog:
## Summary
- `jf gradle artifactorypublish -Ddeploy.test.property='test test'` uploaded the property value to Artifactory with the surrounding single quotes included (`'test test'` instead of `test test`), causing downstream builds to fail when checking that property.
- Root cause: `quotePropertyIfNeeded()` wrapped `-D`/`-P` property values containing spaces in literal single quotes before building the `exec.Command` argv in `gradleRunConfig.GetCmd()`. Since `exec.Command` execs Gradle directly without a shell, those quote characters are never stripped — they're passed straight through to Gradle as part of the value.
- Fix: quoting is now applied only to the string used for the human-readable "Running gradle command" log line, never to the actual arguments passed to `exec.Command`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./build/... -run 'TestQuoteArgsForLog|TestGetCmdDoesNotQuotePropertyValues|TestParseGradleVersion'` — updated existing test (renamed to `TestQuoteArgsForLog`, now testing the log-only helper) and added `TestGetCmdDoesNotQuotePropertyValues`, which asserts `cmd.Args` from `GetCmd()` contains the raw unquoted property value.

Screenshots:

Screenshot 2026-08-25 at 11 15 36 AM

Before fix:
Screenshot 2026-08-25 at 11 19 19 AM

After fix:
Screenshot 2026-08-25 at 11 16 38 AM

### DIFF
--- a/build/gradle.go
+++ b/build/gradle.go
@@ -296,14 +296,7 @@ func unquoteProperty(task string) string {
 	if len(parts) < 2 {
 		return task
 	}
-	value := parts[1]
-	if len(value) >= 2 {
-		first, last := value[0], value[len(value)-1]
-		if (first == '\'' && last == '\'') || (first == '"' && last == '"') {
-			value = value[1 : len(value)-1]
-		}
-	}
-	return parts[0] + "=" + value
+	return parts[0] + "=" + utils.StripSurroundingQuotes(parts[1])
 }
 
 // quoteArgsForLog returns a copy of args with system/project property values wrapped in quotes when they contain
--- a/build/gradle.go
+++ b/build/gradle.go
@@ -260,26 +260,27 @@ func (config *gradleRunConfig) GetCmd() *exec.Cmd {
 	}
 	// Add BUILDINFO_PROPFILE system property if extractor properties file exists
 	if config.extractorPropsFile != "" {
-		jvmProp := fmt.Sprintf("-D%s=%s", extractorPropsDir, config.extractorPropsFile)
-		if strings.Contains(config.extractorPropsFile, " ") {
-			jvmProp = fmt.Sprintf("-D%s='%s'", extractorPropsDir, config.extractorPropsFile)
-		}
-		cmd = append(cmd, jvmProp)
+		cmd = append(cmd, fmt.Sprintf("-D%s=%s", extractorPropsDir, config.extractorPropsFile))
 	}
-	cmd = append(cmd, formatCommandProperties(config.tasks)...)
-	config.logger.Info("Running gradle command:", strings.Join(cmd, " "))
+	cmd = append(cmd, config.tasks...)
+	config.logger.Info("Running gradle command:", strings.Join(quoteArgsForLog(cmd), " "))
 	return exec.Command(cmd[0], cmd[1:]...)
 }
 
-func formatCommandProperties(tasks []string) []string {
-	var cmdArgs []string
-	for _, task := range tasks {
-		if isSystemOrProjectProperty(task) {
-			task = quotePropertyIfNeeded(task)
+// quoteArgsForLog returns a copy of args with system/project property values wrapped in quotes when they contain
+// spaces, e.g., -Dkey=val ue => -Dkey='val ue', purely to make the printed command readable.
+// This must never be used to build the actual arguments passed to exec.Command: those are executed directly,
+// without going through a shell, so added quote characters would be passed through literally instead of being
+// stripped, corrupting the property value.
+func quoteArgsForLog(args []string) []string {
+	logArgs := make([]string, len(args))
+	for i, arg := range args {
+		if isSystemOrProjectProperty(arg) {
+			arg = quotePropertyIfNeeded(arg)
 		}
-		cmdArgs = append(cmdArgs, task)
+		logArgs[i] = arg
 	}
-	return cmdArgs
+	return logArgs
 }
 
 func isSystemOrProjectProperty(task string) bool {
--- a/build/gradle.go
+++ b/build/gradle.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"github.com/jfrog/build-info-go/utils"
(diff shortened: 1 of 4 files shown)


naveenku-jfrog:
Follow-up commit added: not re-adding quotes wasn't sufficient on its own, since on Windows `cmd.exe`/PowerShell don't strip single quotes from arguments the way bash/zsh do. A quoted value can still reach this process with the quotes literally part of the string.

Added `stripPropertyQuotes()`, which strips a matching leading/trailing quote pair (single or double) from `-D`/`-P` property values before they're passed to Gradle, so the property is uploaded clean regardless of shell.

build/gradle.go
```diff
@@ -260,26 +260,56 @@ func (config *gradleRunConfig) GetCmd() *exec.Cmd {
...
+	config.logger.Info("Running gradle command:", strings.Join(quoteArgsForLog(cmd), " "))
 	return exec.Command(cmd[0], cmd[1:]...)
 }
 
-func formatCommandProperties(tasks []string) []string {
-	var cmdArgs []string
-	for _, task := range tasks {
+// stripPropertyQuotes removes a matching pair of leading/trailing quote characters (' or ") from system/project
+// property values, e.g., -Dkey='val ue' => -Dkey=val ue.
+// On Windows, cmd.exe and PowerShell don't strip single quotes from arguments the way bash/zsh do, so a value
```
bhanurp: if this logic applicable to only Windows? If so why not OS check?

build/gradle.go
```diff
@@ -260,26 +260,56 @@ func (config *gradleRunConfig) GetCmd() *exec.Cmd {
...
-			task = quotePropertyIfNeeded(task)
+			task = unquoteProperty(task)
+		}
+		strippedTasks[i] = task
+	}
+	return strippedTasks
+}
+
+// Strips a matching pair of leading/trailing single or double quotes from a system or project property's value.
+func unquoteProperty(task string) string {
```
bhanurp, line 294: move this function to common/utils

build/gradle.go
```diff
@@ -260,26 +260,56 @@ func (config *gradleRunConfig) GetCmd() *exec.Cmd {
...
+	}
+	return parts[0] + "=" + value
+}
+
+// quoteArgsForLog returns a copy of args with system/project property values wrapped in quotes when they contain
+// spaces, e.g., -Dkey=val ue => -Dkey='val ue', purely to make the printed command readable.
+// This must never be used to build the actual arguments passed to exec.Command: those are executed directly,
+// without going through a shell, so added quote characters would be passed through literally instead of being
+// stripped, corrupting the property value.
+func quoteArgsForLog(args []string) []string {
```
bhanurp, line 307: why is this needed? If logged before stripping it should be good enough right?

bhanurp, line 307: missing unit tests for this function

build/gradle.go
```diff
@@ -260,26 +260,56 @@ func (config *gradleRunConfig) GetCmd() *exec.Cmd {
...
+		}
+		strippedTasks[i] = task
+	}
+	return strippedTasks
+}
+
+// Strips a matching pair of leading/trailing single or double quotes from a system or project property's value.
+func unquoteProperty(task string) string {
+	parts := strings.SplitN(task, "=", 2)
+	value := parts[1]
```
fluxxBot: don't you think accessing the array property directly will result in panic if there's no '=' in task

build/gradle.go
naveenku-jfrog: Good catch, done in 3453d82 — stripPropertyQuotes is now gated behind runtime.GOOS == "windows". On other platforms the shell has already stripped shell-level quoting before we see the argument, so any quotes still present were typed deliberately as part of the value; stripping them unconditionally could have silently corrupted a value like -Dkey="'value'" on Unix. Tests updated to assert the OS-specific behavior.

build/gradle.go
naveenku-jfrog: Logging before stripping wouldn't actually reproduce the readability behavior we want in the common (non-Windows) case: on Unix the shell has already stripped any quotes by the time we see config.tasks, so a value like "test test" arrives with no quotes at all, before or after stripPropertyQuotes runs. quoteArgsForLog's job is independent of that — it's purely to wrap space-containing values in quotes for the printed log line so the property boundary is visually clear, regardless of whether the raw arg had quotes. So it's needed either way; keeping it as a separate step (used only for the log string, never for the actual exec.Command args) is what keeps that log-only quoting from leaking into the real argv again, which was the original bug this PR fixes.

naveenku-jfrog: This is already covered — TestQuoteArgsForLog (gradle_test.go) tests quoteArgsForLog directly, including the space-in-value case. Let me know if you had a specific scenario in mind that's not covered.

build/gradle.go
naveenku-jfrog: Fixed in 3453d82 — added a len(parts) < 2 guard in unquoteProperty that returns the task unchanged instead of indexing parts[1]. In practice this was already unreachable, since stripPropertyQuotes only calls unquoteProperty when isSystemOrProjectProperty(task) is true, which itself requires strings.Contains(task, "=") — but making the function safe on its own is better than relying on the caller's guard. Added TestUnquoteProperty covering the no-'=' case directly.

build/gradle.go
naveenku-jfrog: Actually implemented this in 741d054 — pulled the quote-stripping logic out into utils.StripSurroundingQuotes(value string) string (utils/utils.go). unquoteProperty in gradle.go now just handles the key=value split and delegates the quote-stripping to the shared helper. Added TestStripSurroundingQuotes in the new utils/utils_test.go.

fluxxBot (approved):

build/gradle.go
bhanurp: why not move this function unquoteProperty as well this doesnt look like specific to gradle

build/gradle.go
bhanurp: isnt this overkill just for once in a while kind of properties? why not add quotes for all of the property values instad of checking or completely removing?

bhanurp (changes requested):

build/gradle.go
naveenku-jfrog: Good point — simplified in 3ac0220. quoteProperty (renamed from quotePropertyIfNeeded) now always wraps the value in quotes for the log line, dropping the space check entirely. Tests updated.

build/gradle.go
naveenku-jfrog: unquoteProperty itself is tied specifically to Gradle's -D/-P flag convention: it only runs after isSystemOrProjectProperty(task) confirms the arg starts with the systemPropertiesFlag/projectPropertiesFlag constants defined in this file, and no other build tool in this repo uses that -D/-P key=value pattern (checked with a repo-wide grep). What I moved to utils.StripSurroundingQuotes is the generic part — stripping a matching quote pair from an arbitrary string — which has no Gradle dependency. Keeping the -D/-P parsing here avoids pulling Gradle-specific concepts into the shared utils package; happy to revisit if another build tool ends up needing the same -D/-P handling.

bhanurp: What about maven? cant this be reused in maven?

bhanurp (approved):

build/gradle.go
naveenku-jfrog: Fair question — structurally maven.go's GetCmd (mvnRunConfig, build/maven.go:316) has the same shape: a flat []string of -D properties appended straight into exec.Command with no shell, so the same Windows quote-leak could theoretically happen there too. That said, Maven's GetCmd currently does zero quote handling (it never had the original wrap-in-quotes bug either), and I don't have a reported case or repro of a quote-wrapped property actually reaching mvnRunConfig.goals on Windows the way this ticket (RTECO-1885) reported for Gradle's artifactoryPublish. Rather than bolt speculative handling onto a different build tool's file as part of this fix, I'd suggest tracking Maven as a follow-up if/when it's actually hit — utils.StripSurroundingQuotes is already extracted and would be directly reusable there without needing another refactor.

PR #418, opened 2026-08-30
RTECO-1362 - Add strict mode for npm build-info dependency resolution

udaykb2:
## Summary
Add a granular `failOnUncollectedDeps` policy to npm build-info collection. When set, collection fails if a dependency’s integrity/checksum cannot be included in build-info, instead of only logging and publishing a partial module.
- New `CalculateNpmDependenciesList` argument (not part of `NpmTreeDepListParam`): `""` (never fail), `"all"`, or a comma-separated list of `regular`, `peer`, `optional`, `bundle`.
- `handleMissingDeps` applies that policy to all four buckets. Errors are joined so the user sees every missing type at once.
- Without the flag, behavior is unchanged: command succeeds, partial build-info is saved. Peer/bundle still Debug via `printMissingDependenciesWarning`. Optional is Debug and regular is Warn, both using the legacy “will not be included … missing in the npm cache” wording plus the `node_modules` / `package-lock.json` hint (optional’s old “npm ls did not return integrity” text was wrong — optional fails on cache/checksum, same as regular).
- With the flag, optional/regular errors use that same cache sentence; peer/bundle errors still say `npm ls` did not return integrity.

-----

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ bi.exe
 
 # Gradle
 .gradle
+.worktrees
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,3 @@ bi.exe
 
 # Gradle
 .gradle
-.worktrees
--- a/build/utils/npm.go
+++ b/build/utils/npm.go
@@ -23,6 +23,24 @@ import (
 
 const npmInstallCommand = "install"
 
+// Granular values accepted by the --fail-on-missing-deps flag (NpmTreeDepListParam.FailOnMissingDeps).
+// The flag also accepts a comma-separated combination of the granular values, e.g. "peer,optional,bundle".
+const (
+	failOnMissingDepsAll      = "all"
+	failOnMissingDepsRegular  = "regular"
+	failOnMissingDepsPeer     = "peer"
+	failOnMissingDepsOptional = "optional"
+	failOnMissingDepsBundle   = "bundle"
+)
+
+// Dependency type identifiers used internally in handleMissingDeps
+const (
+	depTypePeer     = "peerDependency"
+	depTypeBundle   = "bundleDependencies"
+	depTypeOptional = "optionalDependencies"
+	depTypeRegular  = "regular"
+)
+
 // CalculateNpmDependenciesList gets an npm project's dependencies.
 func CalculateNpmDependenciesList(executablePath, srcPath, moduleId string, npmParams NpmTreeDepListParam, calculateChecksums bool, log utils.Log) ([]entities.Dependency, error) {
 	if log == nil {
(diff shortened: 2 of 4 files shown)


build/utils/npm.go
```diff
@@ -75,17 +93,25 @@ func CalculateNpmDependenciesList(executablePath, srcPath, moduleId string, npmP
...
-	if len(missingOptionalDeps) > 0 {
-		printMissingDependenciesWarning("optionalDependencies", missingOptionalDeps, log)
+	if err := handleMissingDeps(depTypeRegular, otherMissingDeps, npmParams.FailOnMissingDeps, log); err != nil {
+		allErrors = append(allErrors, err.Error())
 	}
-	if len(otherMissingDeps) > 0 {
-		log.Warn("The following dependencies will not be included in the build-info, because they are missing in the npm cache: '" + strings.Join(otherMissingDeps, ",") + "'.\nHint: Try deleting 'node_modules' and/or 'package-lock.json'.")
+	if len(allErrors) > 0 {
+		errorMsg := "Build-info collection stopped.\n" + strings.Join(allErrors, "\n")
+		return nil, errors.New(errorMsg)
```
fluxxBot, line 125: use errorutils.CheckErrorf

build/utils/npm.go
```diff
@@ -266,6 +353,11 @@ type NpmTreeDepListParam struct {
 	IgnoreNodeModules bool
 	// Rewrite package-lock.json, if exists.
 	OverwritePackageLock bool
+	// Fail the build if a dependency's tarball can't be resolved from the npm cache.
+	// Granular string value: "" (never fail, default), "all" (fail for every missing dependency type),
+	// or a comma-separated combination of "regular", "peer", "optional", "bundle"
+	// (e.g. "peer,optional,bundle") to fail only for the specified types.
+	FailOnMissingDeps string
```
fluxxBot: why is this value present here?
I don't think this is the correct struct for this value

fluxxBot (approved):

build/utils/npm.go
```diff
@@ -23,6 +23,24 @@ import (
 
 const npmInstallCommand = "install"
 
+// Granular values accepted by the --fail-on-missing-deps flag (NpmTreeDepListParam.FailOnMissingDeps).
+// The flag also accepts a comma-separated combination of the granular values, e.g. "peer,optional,bundle".
+const (
```
bhanurp, line 28: it should be a boolean flag not string flag, means if it is not failing to identify dependency from npm cache should fail it should n't take these string values.

.gitignore
```diff
@@ -17,3 +17,4 @@ bi.exe
 
 # Gradle
 .gradle
+.worktrees
```
bhanurp: is it required?

build/utils/npm.go
udaykb2: errorutils.CheckErrorf lives in jfrog-client-go, but jfrog-client-go depends on build-info-go, not the other way around — importing it here would be a circular module dependency and wouldn't compile. Keeping errors.New, consistent with how every other error in this file is constructed.

build/utils/npm.go
udaykb2: Moved it out into its own CalculateNpmDependenciesList parameter (now failOnUncollectedDeps

build/utils/npm.go
udaykb2: it lets a caller fail only on the dependency types they actually care about (e.g. peer,optional) rather than an all-or-nothing switch, which was the point of adding this

bhanurp: I am still not sure how does this help user, it has to be all or nothing this is a flag which is making the process strict means it should definetly collect all deps, all being default value says it.

bhanurp (approved):

PR #419, opened 2026-09-01
RTECO-1776 - Implemented debian

fluxxBot:
-----


flexpack/apt/deb822.go
```diff
@@ -0,0 +1,49 @@
...
+			key := line[:idx]
+			value := strings.TrimSpace(line[idx+1:])
+			current[key] = value
+		}
+	}
+
+	// Flush final stanza (file may not end with a blank line)
+	if len(current) > 0 {
+		stanzas = append(stanzas, current)
+	}
```
agrasth, line 46: This loop never checks `scanner.Err()`, and `fillChecksumsFromReader` (the caller in apt_flexpack.go) always returns a `nil` error regardless of what happened here. 
`bufio.Scanner` stops silently - without panicking - if the underlying reader fails (e.g. the `apt-helper cat-file` subprocess in `openPackagesIndex` gets killed or its pipe breaks) or if a single line exceeds the 1 MiB buffer set on line 18. 
Rn either case would just silently resolve fewer checksums, with nothing beyond a lower count in a debug log to explain why - a corrupted or truncated Packages index fails silently instead of surfacing an error.

agrasth (commented):
.

flexpack/apt/deb822.go
fluxxBot: take care

agrasth (approved):

PR #422, opened 2026-09-06
RTECO-1782: warn on externally-resolved deps and dedupe requestedBy paths

bhanurp:
## What

Two dotnet FlexPack build-info fixes, both found during the RTECO-1782 bug hunt.

### 1. Warn when a dependency is resolved outside Artifactory

NuGet records the source it restored each package from in `.nupkg.metadata`. When that source is a **local folder** rather than the injected Artifactory feed, the package never reaches Artifactory — so it is neither curated nor scanned, and gets no repository path in build-info.

This happens routinely and silently. Every F# project resolves `FSharp.Core` from the SDK's `FSharp/library-packs`:

```jsonc
// fsharp.core/10.1.302/.nupkg.metadata
"source": "/usr/local/share/dotnet/sdk/10.0.302/FSharp/library-packs"

// newtonsoft.json/13.0.3/.nupkg.metadata
"source": "https://acme.jfrog.io/artifactory/api/nuget/v3/nuget-virtual/index.json"
```

The UI showed `No path found (externally resolved or deleted/overwritten)` with no explanation. The collector's checksum was correct all along — it hashed the real resolved file. What was missing was any signal that the package bypassed Artifactory.

Now one aggregated warning names them:

```
[Warn] 1 dependency was resolved outside Artifactory and therefore not curated or scanned:
  FSharp.Core:10.1.302 (source: /usr/local/share/dotnet/sdk/10.0.302/FSharp/library-packs)
  ... build with -p:DisableImplicitLibraryPacksFolder=true (F# projects) or clear the fallback folder.
```

**Scoping:** only non-HTTP sources are reported. This layer doesn't know which feed URL was injected, so flagging remote URLs would risk false positives; a local folder is unambiguous.

### 2. Deduplicate `requestedBy` paths by immediate parent

Xray's SBOM builder reads **only `path[0]`** of each path — the immediate parent — and reassembles the tree from every package's own entry. Paths sharing a `path[0]` therefore carry no extra information, but they *do* consume the `RequestedByMaxLength` (15) budget. That can push out a parent recorded nowhere else and **silently drop an SBOM edge**.

Measured on a sample solution, before and after:

```
before: 58 paths emitted, 33 distinct parents  -> 25 redundant
after:  33 paths emitted, 33 distinct parents  ->  0 redundant
```

43% smaller payload, **identical edge set**. `Microsoft.Extensions.DependencyInjection.Abstractions` was sitting on exactly 15 paths for 6 distinct parents — at the cap, with 9 slots wasted.

## Invariant worth reviewing

A direct dependency's single-element `{moduleId}` path is what attaches it to the SBOM graph **root** — Xray builds its id→entry map from the module id plus every dependency id. Strip it and direct dependencies become orphan nodes. Being the shortest possible path it always wins its parent slot, and three regression tests pin that, including the adversarial case where a longer path shares the module as its `path[0]`.

## ⚠️ One existing test expectation changed

`TestPopulateRequestedByDeterministic` expected two paths for `depD`:

```go
{"depC:1", "depA:1", "TestModule"}
{"depC:1", "depB:1", "TestModule"}
```

Both share `path[0] = depC:1` and so produce the **same** edge `depC→depD`. One is kept now. The depA/depB distinction is not lost — it lives on `depC`'s own entry, where it genuinely carries two edges.

Please sanity-check this one: changing a test to match a behaviour change deserves a second pair of eyes.

## Testing

- `TestLocalResolutionSource` — 8 cases, including that missing/malformed metadata is **not** reported (absence of evidence is not a bypass) and that path traversal is rejected
- `TestDedupeRequestedByPreservesDistinctParents` — 7 cases covering the direct-dependency invariant above
- `gofmt`, `go build ./...`, `go vet`, `golangci-lint` — all clean; full `build/utils/dotnet/...` suite passes
- Verified end-to-end against a live Artifactory across 8 project types (csproj, fsproj, vbproj, sln, slnx, CPM, packable lib, lockfile)

> **Note:** `gosec` could not be run — it fails with `internal error: package "encoding/json" without types` under this Go toolchain. That check is unverified rather than passing.

## Merge order

This is the **first** of three RTECO-1782 PRs. `jfrog-cli` needs this merged and bumped, because `dotnet_native_test.go` asserts the dedupe behaviour.

1. **build-info-go** ← this PR
2. jfrog-cli-artifactory
3. jfrog-cli

### DIFF
--- a/build/utils/dotnet/dependencies/assetsjson.go
+++ b/build/utils/dotnet/dependencies/assetsjson.go
@@ -151,12 +151,18 @@ func (assets *assets) getAllDependencies(log utils.Log) (map[string]*buildinfo.D
 	// NUGET_PACKAGES path or the SDK fallback folder); checksums are added when the file is
 	// available. This prevents dependencies from being silently dropped (jfrog-cli#600, #1796).
 	privateDeps := assets.getPrivateDependencyNames()
+	var externallyResolved []string
 	for dependencyId, library := range assets.Libraries {
 		if library.Type == "project" {
 			continue
 		}
 		dependencyKey := strings.ToLower(getDependencyIdForBuildInfo(dependencyId))
 		dependency := &buildinfo.Dependency{Id: getDependencyIdForBuildInfo(dependencyId), Type: nupkgType}
+
+		if source := localResolutionSource(packagesPath, library); source != "" {
+			externallyResolved = append(externallyResolved,
+				fmt.Sprintf("%s (source: %s)", getDependencyIdForBuildInfo(dependencyId), source))
+		}
 
 		checksum, err := assets.dependencyChecksum(packagesPath, library, log)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


bhanurp:
Superseded by #424, which is the same work on a branch in this repository rather than a fork. A PR's head branch cannot be changed after creation, so reopening was the only way to move it.

PR #425, opened 2026-09-10
RTECO-2003: add Chocolatey FlexPack collectors

bhanurp:
## Summary

Adds `flexpack/choco`: artifact collection for `choco pack` / `choco push`, and dependency collection for `choco install` / `upgrade`. Chocolatey writes no lock file, so the transitive graph is recovered from each installed package's `.nuspec`.

Part of RTECO-2003 (`jf choco` + `jf setup choco`). Ships first; the other three PRs pin this commit.

## Three native behaviours worth calling out

Each was verified against Chocolatey's documentation or its own source rather than assumed, and each fails **silently** if modelled wrongly:

| Behaviour | Consequence if missed |
|---|---|
| `choco pack` accepts `--out` / `--outdir` / `--outputdirectory` / `--output-directory` | Snapshotting only the working directory finds nothing the pack produced → empty build-info, no error |
| An installed package is `lib/<id>/<id>.nupkg` — **no version in the file name** | Reading the version off the name always yields `""` → every dependency dropped |
| `choco push` takes an **optional** path, pushing the single `.nupkg` in the folder | A bare push uploads the package but records no artifact → unstamped, absent from build-info |

The layout claim is verifiable in Chocolatey's own integration suite, which asserts paths of the form `lib/isdependency/isdependency.nupkg`.

Also treats a Chocolatey 403 as forbidden output (reusing the NuGet wording) so curation-on-failure classifies a blocked package correctly.

## Test plan

- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./flexpack/choco/... ./utils/...`
- [x] New unit tests cover all three behaviours above, plus the ambiguous-folder case (records nothing rather than guessing) and the API-key spellings that must never be mistaken for a package path

### DIFF
--- a/flexpack/choco/choco_artifacts.go
+++ b/flexpack/choco/choco_artifacts.go
@@ -1,0 +1,306 @@
+package choco
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/jfrog/build-info-go/entities"
+	"github.com/jfrog/gofrog/crypto"
+)
+
+const (
+	nupkgExtension = ".nupkg"
+	nupkgType      = "nupkg"
+)
+
+type snapshotEntry struct {
+	modTime time.Time
+	sha1    string
+}
+
+// PackageSnapshot records .nupkg files in a Chocolatey pack working directory.
+type PackageSnapshot map[string]snapshotEntry
+
+// SnapshotPackageFiles records the .nupkg files in root and in each of extraDirs. extraDirs carries
+// the directory named by `choco pack --output-directory` (and its aliases): that flag exists, and
+// without it a pack into a custom directory is invisible to the snapshot diff, so the build-info
+// comes back with no artifacts and no error at all.
+func SnapshotPackageFiles(root string, extraDirs ...string) (PackageSnapshot, error) {
+	if root == "" {
+		return nil, errors.New("the Chocolatey package snapshot root must not be empty")
+	}
+	snapshot := make(PackageSnapshot)
+	for _, directory := range append([]string{root}, extraDirs...) {
+		if err := addDirectoryToSnapshot(snapshot, directory, directory == root); err != nil {
+			return nil, err
+		}
+	}
+	return snapshot, nil
+}
+
+// addDirectoryToSnapshot indexes one directory. A missing extra directory is not an error: choco
+// creates the output directory itself, so it legitimately may not exist when the snapshot is taken.
+func addDirectoryToSnapshot(snapshot PackageSnapshot, directory string, required bool) error {
+	if directory == "" {
+		return nil
+	}
+	entries, err := os.ReadDir(directory)
+	if err != nil {
+		if !required && os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("read Chocolatey package directory %s: %w", directory, err)
+	}
+	for _, entry := range entries {
+		if entry.IsDir() || !isNupkg(entry.Name()) {
+			continue
+		}
+		path := filepath.Join(directory, entry.Name())
+		info, err := entry.Info()
+		if err != nil {
+			return fmt.Errorf("stat Chocolatey package %s: %w", path, err)
+		}
+		details, err := crypto.GetFileDetails(path, true)
+		if err != nil {
+			return fmt.Errorf("compute snapshot checksum for %s: %w", path, err)
+		}
+		absolutePath, err := filepath.Abs(path)
+		if err != nil {
+			return fmt.Errorf("resolve Chocolatey package path %s: %w", path, err)
+		}
+		snapshot[absolutePath] = snapshotEntry{modTime: info.ModTime(), sha1: details.Checksum.Sha1}
+	}
+	return nil
+}
+
+// ExtractPackOutputDirectory returns the directory named by `choco pack`'s output-directory flag,
+// or "" when the command did not pass one. Chocolatey accepts four spellings, in both the
+// "--flag value" and "--flag=value" forms.
+func ExtractPackOutputDirectory(args []string) string {
+	for index := 0; index < len(args); index++ {
+		arg := args[index]
+		if !strings.HasPrefix(arg, "-") {
+			continue
+		}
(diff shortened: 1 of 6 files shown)


flexpack/choco/choco_flexpack.go
```diff
@@ -0,0 +1,420 @@
...
+				}
+				continue
+			}
+			return nil, nil, nil, err
+		}
+		dependencyID := packageName + ":" + version
+		requester := pending.requestedBy[0]
+		graph[requester] = appendUniqueEdge(graph[requester], dependencyID)
+
+		if existing, seen := visited[strings.ToLower(packageName)]; seen {
```
agrasth, line 157: When the dependency walker hits a package it's already seen — because two other packages both depend on it (a completely normal situation called a **"diamond" dependency**) — it records the new path here.
However, this same code path also fires when the walker hits a **cycle** (package A depends on B, and B depends back on A).

In that case, it does the exact same thing: it tacks the new path onto the already-visited package's own `RequestedBy` list. Except this time, that "new path" contains the package's own ID.

The code comment even says:

> "this is also what terminates a dependency cycle"

treating the cycle case as if it's just the same as the diamond case — but it isn't.

- A **diamond dependency** needs the extra path recorded.
- A **cycle** needs that specific edge dropped instead, because you're about to list a package as being requested by a chain that includes itself.

flexpack/choco/choco_test.go
```diff
@@ -0,0 +1,378 @@
...
+
+func TestBuildArtifactModules(t *testing.T) {
+	modules := BuildArtifactModules([]entities.Artifact{{Name: "git.install.2.43.0.2.nupkg", Type: nupkgType}}, "")
+	require.Len(t, modules, 1)
+	assert.Equal(t, "git.install:2.43.0.2", modules[0].Id)
+}
+
+// Chocolatey installs the fully resolved dependency set under lib, so the transitive graph is
+// recovered from the .nuspec of each installed package rather than from a lock file.
+func TestCollectTransitiveDependencies(t *testing.T) {
```
agrasth, line 101: This test builds exactly this shape (`tool → libA → tool` cycle) and carefully asserts `RequestedBy` for `libA:1.1.0` (line 150) and `libShared:2.0.0` (line 153).
However, it never checks `byID["tool:1.0.0"].RequestedBy`, which is the one value that would expose the corruption.

It does verify that the dependency graph terminates correctly (line 160: *"a dependency cycle is reported as an edge, and must not loop forever"*), so **termination was tested**, but the **content of the cyclic package's own `RequestedBy` was not**.

We can try adding a check before appending, something along the lines of:

> Skip appending `pending.requestedBy` to `existing.requestedBy` when `pending.requestedBy` already contains `existing`'s own dependency ID.

In other words, don't allow a package's own ID to appear inside its own `RequestedBy` chain.
The `apt` collector's `onPath`-set approach in this same repository is a working reference for the appropriate shape of the fix.

flexpack/choco/choco_artifacts.go
```diff
@@ -0,0 +1,306 @@
...
+	default:
+		return false
+	}
+}
+
+func isNupkg(name string) bool {
+	return strings.HasSuffix(strings.ToLower(name), nupkgExtension)
+}
+
+func parseNupkgFilename(filename string) (string, string) {
```
agrasth, line 289: Nit: function `parseNupkgFilename` (used only for grouping pack/push artifacts into a module, not for the install/upgrade path — that one correctly reads the real version from `.nuspec`) — its heuristic for splitting `packageid.version.nupkg` assumes the first dot-segment after the first one that starts with a digit is where the version begins. 
For a package ID with a numeric-looking segment in the name itself (e.g. `foo.2024.crawler.1.0.0.nupkg`), it would misparse the ID/version split and mis-group the artifact's module. Narrow real-world impact (display/grouping only, never affects the recorded checksum or path)

flexpack/choco/choco_flexpack.go
bhanurp: Fixed in 554e35b: the walk now checks whether the revisiting chain already contains the package's own id before appending it to `RequestedBy`, so a cycle edge (A -> B -> A) is dropped instead of being recorded as a self-referential path. Diamond dependencies (a genuinely different requester) are unaffected.

flexpack/choco/choco_test.go
bhanurp: Added in 554e35b: `byID["tool:1.0.0"].RequestedBy` is now asserted to equal `[["image-build"]]` - verified this assertion fails (showing the corrupted self-referential chain) without the cycle fix, and passes with it.

flexpack/choco/choco_artifacts.go
bhanurp: Left as-is: this is the identical heuristic already in flexpack/nuget/nuget_artifacts.go ("NuGet convention: the first dot-separated segment beginning with a digit starts the version"), so changing it here alone would diverge from an established, intentional convention rather than fix a choco-specific regression. Agreed the edge case is real but, as noted, narrow and display/grouping-only. Leaving open for a follow-up covering both collectors together if we want to harden this.

fluxxBot (approved):

flexpack/choco/choco_flexpack.go
```diff
@@ -0,0 +1,440 @@
...
```
agrasth: One more thing worth adding while we're in here: this loop has no cap on how many paths pile up in `existing.requestedBy`. 
The apt and dotnet collectors in this same repo both stop adding paths once a dependency hits `entities.RequestedByMaxLength (15)` — this one doesn't. For a package that's a dependency of many other installed packages (a common runtime lib, say), its RequestedBy list could just keep growing with no limit.

agrasth (approved):

flexpack/choco/choco_flexpack.go
bhanurp: Good catch - fixed in de66911: existing.requestedBy is now also capped at entities.RequestedByMaxLength before appending, matching the apt/dotnet collectors' convention. Added TestCollectDependencyTreeCapsRequestedBy (20 requesters of a shared leaf package) to pin it - verified the assertion fails at 20 entries without the cap and passes at 15 with it.

PR #430, opened 2026-09-21
RTECO-2247: PSResource (PowerShell PSResourceGet) FlexPack collectors

bhanurp:
## Summary

Part of RTECO-2247 — adds FlexPack build-info collection support for PowerShell's PSResourceGet module, mirroring the existing `flexpack/choco` collector's design. This is one of four repos in this change; see the sibling PRs for the full picture:

- jfrog-cli-core: adds `project.PSResource` (link added once opened)
- jfrog-cli-artifactory: `jf setup psresource` + the FlexPack command that calls into this package (link added once opened)
- jfrog-cli: registers `jf Install-/Save-/Update-/Publish-PSResource` as top-level commands (link added once opened)

## What's here

- `flexpack/psresource/psresource_flexpack.go` — `PSResourceFlexPack`, a `BuildInfoCollector` implementation for `Install-/Save-/Update-PSResource`. Turns already-resolved dependency data (name/version/checksum) into `entities.Dependency` records.
- `flexpack/psresource/psresource_artifacts.go` — artifact assembly for `Publish-PSResource`, including `DerivePublishedPath` (the NuGet-convention path formula both this package and the caller need to agree on).
- `flexpack/flexpack.go` — `PSResourceConfig` and the shared `ResolvedPackage` type.
- `utils/error.go` — `PSResource` package-manager constant + `IsForbiddenOutput` case.

**Design note**: this package does zero network or filesystem I/O, matching every other collector here. PSResourceGet has no lock file and `Publish-PSResource` uploads directly without leaving a local `.nupkg`, so the caller (jfrog-cli-artifactory, which has the Artifactory server/auth context) is responsible for resolving package identity (`Get-InstalledPSResource`) and checksums (a HEAD request to Artifactory) and handing the results in via `PSResourceConfig.ResolvedPackages`/`PublishedArtifact`.

## Test plan

- `go test ./flexpack/psresource/... ./utils/...` — all pass, no regressions in adjacent packages.
- New coverage: `utils/error_test.go` (this repo had no `error_test.go` at all) covering `IsForbiddenOutput`'s `psresource` branch; `psresource_test.go` covering `CollectBuildInfo`'s error-propagation path in addition to the happy paths.
- `gofmt`/`go vet`/`golangci-lint` clean on all touched files.

- [x] All tests passed.
- [x] All static analysis checks passed.
- [x] This pull request is on the main branch (the repo's actual default branch — `dev` is 163 commits behind and appears stale).


flexpack/psresource/psresource_artifacts.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Repo is the target Artifactory repository the package was published to.
+	Repo string
+}
+
+// BuildPublishedArtifact assembles an entities.Artifact for a Publish-PSResource command from the
+// already-resolved package data in published. It performs no network or filesystem I/O: checksum
+// resolution (via a HEAD request to Artifactory) is the caller's responsibility.
+func BuildPublishedArtifact(published PublishedArtifact) (entities.Artifact, error) {
+	if err := validateNameVersion("published", published.Name, published.Version); err != nil {
```
naveenku-jfrog: "published" should be constant.

flexpack/psresource/psresource_flexpack.go
```diff
@@ -0,0 +1,102 @@
...
+
+// compile-time assertion that PSResourceFlexPack satisfies the same interface every other FlexPack
+// collector in this repo does.
+var _ buildinfoflex.BuildInfoCollector = (*PSResourceFlexPack)(nil)
+
+// NewPSResourceFlexPack creates a new PSResourceFlexPack with the given configuration.
+func NewPSResourceFlexPack(config buildinfoflex.PSResourceConfig) (*PSResourceFlexPack, error) {
+	if config.WorkingDirectory == "" {
+		return nil, fmt.Errorf("PSResourceConfig.WorkingDirectory must not be empty")
+	}
```
naveenku-jfrog: If workingDirectory is empty, then can't we take current working Directory?

flexpack/psresource/psresource_flexpack.go
```diff
@@ -0,0 +1,102 @@
...
+// already-resolved packages in config.ResolvedPackages. It performs no network or filesystem I/O.
+func (p *PSResourceFlexPack) CollectBuildInfo(buildName, buildNumber string) (*entities.BuildInfo, error) {
+	dependencies, err := BuildDependencies(p.config.ResolvedPackages)
+	if err != nil {
+		return nil, fmt.Errorf("collect PSResource dependencies: %w", err)
+	}
+
+	moduleID := p.config.Module
+	if moduleID == "" {
+		moduleID = "psresource-project"
```
naveenku-jfrog: why hardcoded string?

flexpack/psresource/psresource_flexpack.go
```diff
@@ -0,0 +1,102 @@
...
+// BuildDependencies turns already-resolved PSResource packages into entities.Dependency records.
+// This is pure data assembly: the caller is responsible for resolving package identity (via
+// Get-InstalledPSResource) and checksums (via a HEAD request to Artifactory) before calling this.
+func BuildDependencies(resolved []ResolvedPackage) ([]entities.Dependency, error) {
+	if len(resolved) == 0 {
+		return nil, nil
+	}
+	dependencies := make([]entities.Dependency, 0, len(resolved))
+	for _, pkg := range resolved {
+		if err := validateNameVersion("resolved", pkg.Name, pkg.Version); err != nil {
```
naveenku-jfrog: Why hardcoded string?

flexpack/psresource/psresource_flexpack.go
```diff
@@ -0,0 +1,102 @@
...
+			return nil, err
+		}
+		dependencies = append(dependencies, entities.Dependency{
+			Id:       nupkgFileName(pkg.Name, pkg.Version),
+			Type:     nupkgType,
+			Scopes:   []string{"main"},
```
naveenku-jfrog: Why manually setting the Scopes as main only?

flexpack/psresource/psresource_flexpack.go
```diff
@@ -0,0 +1,102 @@
...
+// PSResourceConfig.ResolvedPackages, so this always returns an empty result.
+func (p *PSResourceFlexPack) GetProjectDependencies() ([]buildinfoflex.DependencyInfo, error) {
+	return nil, nil
+}
+
+// GetDependencyGraph returns an empty graph since PSResource has no lock file
+// and we only track direct dependencies.
+func (p *PSResourceFlexPack) GetDependencyGraph() (map[string][]string, error) {
+	return make(map[string][]string), nil
+}
```
naveenku-jfrog: These 2 functions are only called inside test files, then why we require these at all?

naveenku-jfrog (changes requested):

flexpack/psresource/psresource_artifacts.go
bhanurp: Good catch, fixed. Hoisted "published" (and the sibling "resolved" literal used in the same validateNameVersion call from psresource_flexpack.go) into named consts `publishedPackageKind` / `resolvedPackageKind` in psresource_artifacts.go.

flexpack/psresource/psresource_flexpack.go
bhanurp: Agreed, this was a real functional gap. Fixed: an empty WorkingDirectory now falls back to os.Getwd() instead of erroring, matching the exact pattern HelmFlexPack already uses (helm_flexpack.go). Updated the existing unit test that asserted the old error-on-empty behavior to assert the os.Getwd() fallback instead.

flexpack/psresource/psresource_flexpack.go
bhanurp: Fixed: hoisted "psresource-project" into a named const defaultModuleID, with a comment noting it mirrors the same <tool>-project fallback convention other FlexPack collectors in this repo use (e.g. cargo-project, ruby-project). No shared constant exists for this across collectors, so a local const is the right scope.

flexpack/psresource/psresource_flexpack.go
bhanurp: Fixed, same treatment as the "published" comment above: this "resolved" literal now uses the named const resolvedPackageKind (defined alongside publishedPackageKind in psresource_artifacts.go where validateNameVersion lives).

flexpack/psresource/psresource_flexpack.go
bhanurp: Checked how the other flat-list-style collectors handle this (e.g. Poetry also hardcodes Scopes: []string{"main"} for one of its dependency shapes; RubyGems/Maven/Gradle derive scopes dynamically because they *do* have dev/optional/test dependency groups). PSResourceGet has no such distinction: Get-InstalledPSResource returns a flat list of installed modules with no group/category metadata to derive a scope from, so "main" is genuinely the only scope that will ever apply here. Kept the behavior, but extracted it into a named const defaultScope with a comment explaining this reasoning so it's not read as an arbitrary hardcode.

flexpack/psresource/psresource_flexpack.go
bhanurp: These two methods (GetProjectDependencies/GetDependencyGraph) are required by the buildinfoflex.BuildInfoCollector interface that every FlexPack collector in this repo implements (see the compile-time assertion var _ buildinfoflex.BuildInfoCollector = (*PSResourceFlexPack)(nil) a few lines above). They're not dead code: the real caller, jfrog-cli-artifactory, invokes them polymorphically through that interface rather than through the concrete *PSResourceFlexPack type, which is exactly why grepping this repo alone only turns up calls from psresource_test.go - the same pattern holds for every other collector here (nix, cargo, poetry, uv, rubygems, nuget, maven all implement these two methods for the same reason). Removing them would break interface conformance and any caller relying on it, so I've left them as-is and added a comment above them explaining why they exist despite the test-only call sites in this repo.

naveenku-jfrog (approved):

